PR #464, opened 2026-04-01
Test/manufacturing

Kanishkar16Vijay:


jewellery_erpnext/jewellery_erpnext/doctype/employee_ir/test_employee_ir.py
```diff
@@ -1,9 +1,285 @@
...
+from jewellery_erpnext.jewellery_erpnext.doctype.manufacturing_operation.test_manufacturing_operation import (
+	dir_for_issue,
+	dir_for_receive,
+	scan_mwo_eir,
+)
+
 
 class TestEmployeeIR(FrappeTestCase):
-	pass
+	def setUp(self):
```
Vigneshsekar, line 145: ### Blocker findings

#### Import mismatch: `book_metal_loss`
- Test file imports:
  - `book_metal_loss`
  - `create_operation_for_next_op`
  - `get_manufacturing_operations`
  - `get_rows_to_append`
- On `develop_aerele`:
  - `create_operation_for_next_op` exists as a module-level function.
  - `get_manufacturing_operations` exists as a module-level function.
  - **`book_metal_loss` is not module-level; it is an instance method on `EmployeeIR`.**
  - **`get_rows_to_append` is not present as a module-level symbol in `employee_ir.py`.**
- That makes the test module import itself invalid against the base branch implementation.

#### `test_book_metal_loss_returns_empty_when_no_difference`
- Calls `book_metal_loss({}, "MWO-1", "OPT-1", 10, 10)` as if it were a free function.
- The real implementation is `EmployeeIR.book_metal_loss(self, mwo, opt, gwt, r_gwt, allowed_loss_percentage=None)`.
- So even conceptually the invocation does not match the source API.

### Major findings

#### `test_get_rows_to_append_returns_rows_for_positive_qty`
- Since `get_rows_to_append` is not available as imported here, this test cannot be valid on the actual source.
- Even ignoring the import issue, it only checks output shape, not whether the rows are later acceptable for the downstream stock-entry path.

#### `test_validate_process_loss_calculates_loss_correctly`
- Calls `validate_process_loss()` and then asserts `len(eir.employee_loss_details) >= 0`.
- That assertion is always true.
- It does **not** validate that loss was calculated, only that the child table object exists.
- This does not meaningfully cover the source method.

#### `test_before_validate_throws_error_when_stock_reconciliation_in_progress`
- Uses a broad `frappe.db.get_value` patch with a fixed side-effect list.
- The test assumes the call order and all call sites inside `before_validate()`, which is brittle.
- It does not assert the actual message or which condition triggered the ValidationError.

### Minor / quality issues

#### `test_department_ir_by_manufacturing_operation`
- For the receive side it conditionally mutates `eir_issue.employee_ir_operations[0].rpt_wt_issue` instead of `eir_receive...`.
- That looks like a setup mistake. It may not fail depending on the environment, but it is clearly the wrong object.

#### `test_get_manufacturing_operations_does_not_duplicate_if_present`
- Valid enough, but only checks count stability. It does not verify that the exact existing row remains untouched.

### Missing coverage
- No test for `create_operation_for_next_op()` clearing child tables and reset fields beyond a few scalar checks.
- No negative-path test for `get_manufacturing_operations()` when called with serialized `target_doc` string, which the source explicitly supports.
- No proper assertion for proportional loss rows generated by `validate_process_loss()`.

Vigneshsekar (changes requested):

jewellery_erpnext/jewellery_erpnext/doctype/manufacturing_operation/test_manufacturing_operation.py
```diff
@@ -1,9 +1,541 @@
...
+	DepartmentIR,
+)
+from jewellery_erpnext.jewellery_erpnext.doctype.manufacturing_operation.manufacturing_operation import (
+	ManufacturingOperation,
+	get_material_wt,
+	get_stock_entries_against_mfg_operation,
+)
+
 
 class TestManufacturingOperation(FrappeTestCase):
```
Vigneshsekar, line 18: ### Blocker findings

#### `test_get_stock_entries_against_mfg_operation_aggregation`
- Real source logic in `get_stock_entries_against_mfg_operation(doc)`:
  1. resolve warehouse from `doc.department`, or `doc.employee`, or subcontractor path
  2. query `Stock Entry Detail` using that resolved `t_warehouse` and `manufacturing_operation`
- The test creates:
  - a bare MOP with no department/company linkage to warehouse resolution
  - a standalone Warehouse not tied to the MOP via department/employee/subcontractor filters used by source
  - a Stock Entry inserted with `docstatus = 1` directly
  - Stock Entry Detail rows inserted directly with `docstatus = 1`
- This setup does **not** mirror the actual lookup path used by the source function.
- So the test is not proving the source behavior it claims to test.

### Major findings

#### `test_validate_loss_raises_on_invalid_item`
- Intended to exercise `validate_loss()`.
- It uses `with self.assertRaises(Exception)` instead of asserting `frappe.ValidationError` or the source message.
- Too broad: it will pass on unrelated setup errors.

#### `test_has_overlap_detects_overlap`
- The first case is fine for overlap.
- The second case uses sequential intervals (`12:00` starts exactly when prior ends) and still expects `True`.
- That expectation matches the current source implementation because `production_capacity == 1 and len(time_logs) >= 1` returns `True` immediately, but it does **not** validate real overlap semantics. It effectively locks in a questionable implementation quirk rather than testing business behavior.

#### `test_manufacturing_operations`
- Very broad end-to-end test.
- It asserts many downstream document linkages, which is useful.
- But it also relies on helper functions defined in the same test module and mutable ambient fixtures, so failures will be hard to localize.

### Missing coverage
- No unit test for `validate_loss()` wrong-UOM path.
- No unit test for `validate_loss()` quantity-greater-than-available path.
- No direct test for `set_start_finish_time()` behavior even though time-log handling is heavily exercised elsewhere.

Vigneshsekar (changes requested):

jewellery_erpnext/jewellery_erpnext/doctype/parent_manufacturing_order/test_parent_manufacturing_order.py
```diff
@@ -1,9 +1,270 @@
...
 from frappe.tests.utils import FrappeTestCase
 
+from jewellery_erpnext.jewellery_erpnext.doctype.parent_manufacturing_order.parent_manufacturing_order import (
+	ParentManufacturingOrder,
+	get_item_code,
+	validate_mfg_date,
+)
+
 
 class TestParentManufacturingOrder(FrappeTestCase):
```
Vigneshsekar, line 15: ### Blocker findings

#### `test_manufacturing_work_order_creation_with_multicolour`
- Uses:
  ```python
  bom.append("metal_detail", [ {...}, {...} ])
  ```
- Frappe child-table `append()` expects a **single row dict**, not a list of dicts.
- This setup is incompatible with the framework API and is not a valid way to create two child rows.
- The test therefore cannot be relied on as written.

### Major findings

#### `test_validate_mfg_date_throws_on_invalid_dates`
- Patches `frappe.throw` to raise plain `Exception`.
- That does exercise the branch, but it stops validating the actual exception type used by source.
- Acceptable as a branch test, but weak as contract validation.

#### `test_create_material_requests_throws_when_no_bom`
- Similar issue: patches `frappe.throw` to plain `Exception`.
- It verifies the branch but not the actual framework error type/message contract.

### Missing coverage
- No direct assertion on per-material-request grouping logic inside `create_material_requests()`.
- No negative-path test for missing manufacturer warehouse configuration, which is a real branch in source.

Vigneshsekar (changes requested):

jewellery_erpnext/jewellery_erpnext/doctype/parent_manufacturing_order/test_parent_manufacturing_order.py
Kanishkar16Vijay: Resolved the changes bro

Kanishkar16Vijay:
@Vigneshsekar bro, kindly review the PR and I have committed the changes

PR #465, opened 2026-04-01
refactor: optimize PMO DocType for performance

manikandan-s-18:


jewellery_erpnext/jewellery_erpnext/doctype/parent_manufacturing_order/parent_manufacturing_order.py
```diff
@@ -195,287 +231,175 @@ def update_estimated_delivery_date_in_prev_docs(self):
...
-					"Variant based Warehouse",
-					{"parent": self.manufacturer},
-					["variant", "department", "target_warehouse"],
-				)
+		for bom_table, fields in _BOM_TABLE_FIELDS.items():
+			data = frappe.get_all(bom_table, {"parent": bom}, fields)
+			if not data:
+				continue
+			for row in data:
+				item_type = get_item_type(row.get("item_variant"))
```
Vigneshsekar: Can row.get("item_variant") return none?

jewellery_erpnext/jewellery_erpnext/doctype/parent_manufacturing_order/parent_manufacturing_order.py
```diff
@@ -195,287 +231,175 @@ def update_estimated_delivery_date_in_prev_docs(self):
...
+				variant_key = _ITEM_TYPE_PREFIX[item_type]
+				if variant_key not in warehouse_dict:
+					frappe.throw(_("Please mention warehouse details in Manufacturer"))
+				dept_info = warehouse_dict[variant_key]
+				from_wh = _get_raw_material_warehouse(dept_info.department)
```
Vigneshsekar: As this is within the loop, instead of _get_raw_material_warehouse(dept_info.department) can you fetch all the from_wh for all the departments and keep it before the loop or put that in cache as a JSON and retrieve it? It most likely may not change, isn't it?

jewellery_erpnext/jewellery_erpnext/doctype/parent_manufacturing_order/parent_manufacturing_order.py
```diff
@@ -773,14 +618,7 @@ def create_manufacturing_work_order(self):
 				):
 					doc.multicolour = 1
 					doc.allowed_colours = color["metal_colours"]
-			doc.metal_touch = row.metal_touch
```
Vigneshsekar, line 776: But doc is coming from mapped docs and row is above it. What if the doc is not having these info? Is it okay to remove it?

Bcoz, these are only assignments. No fn. calls or db. calls. So if it is safe and correct to keep it, please keep it. 

Same for all `metal_touch, metal_type, metal_purity, metal_colour`

Vigneshsekar (changes requested):

jewellery_erpnext/jewellery_erpnext/doctype/parent_manufacturing_order/parent_manufacturing_order.py
manikandan-s-18: Yes, that's correct, explicit assignment is safe in this case. The required assignments are already present in lines 789–792, so I removed the redundant duplicate block to avoid repetition

manikandan-s-18:
@Vigneshsekar changes have been applied.

jewellery_erpnext/jewellery_erpnext/doctype/parent_manufacturing_order/parent_manufacturing_order.py
```diff
@@ -21,43 +19,93 @@
...
+		"pcs",
+	],
+	"BOM Gemstone Detail": [
+		"item_variant",
+		"quantity",
+		"is_customer_item",
+		"sub_setting_type",
+		"pcs",
+	],
+	"BOM Other Detail": ["item_code", "quantity", "qty"],
```
Vigneshsekar, line 55: item_variant is not here

jewellery_erpnext/jewellery_erpnext/doctype/parent_manufacturing_order/parent_manufacturing_order.py
```diff
@@ -195,287 +231,183 @@ def update_estimated_delivery_date_in_prev_docs(self):
...
-				)
-			if data:
-				# Loop through the rows of the bom table
-				variant_warehouse = frappe.db.get_all(
-					"Variant based Warehouse",
-					{"parent": self.manufacturer},
-					["variant", "department", "target_warehouse"],
-				)
+		for bom_table, fields in _BOM_TABLE_FIELDS.items():
+			data = frappe.get_all(bom_table, {"parent": bom}, fields)
```
Vigneshsekar, line 268: So, in the fields, item_variant will not be fetched

jewellery_erpnext/jewellery_erpnext/doctype/parent_manufacturing_order/parent_manufacturing_order.py
```diff
@@ -195,287 +231,183 @@ def update_estimated_delivery_date_in_prev_docs(self):
...
+			if not data:
+				continue
+			for row in data:
+				if not row.get("item_variant"):
```
Vigneshsekar: So, it will be none always, isn't it?

Vigneshsekar (changes requested):

jewellery_erpnext/jewellery_erpnext/doctype/parent_manufacturing_order/parent_manufacturing_order.py
Vigneshsekar: If it returns none, then this DB call with get_item_type is wasted. So, you can check if it is none, if none, you can ignore to proceed.

jewellery_erpnext/jewellery_erpnext/doctype/parent_manufacturing_order/parent_manufacturing_order.py
manikandan-s-18: Yeah,  that is resolved now.

Vigneshsekar:
Ready to merge

Vigneshsekar (approved):

PR #660, opened 2026-06-01
feat: create_custom_field

Praveenkumar26-S:


DHINESH00:
## Review — #660 create_custom_field

&nbsp;·&nbsp; 3 blocker · 3 high · 5 medium · 1 low · 1 nit
&nbsp;·&nbsp; base `v16_develop_aerele` · head `bade320d` · 5 files, +125935/-549

Still a draft, and the branch is roughly 3.5 months behind its base (merge base `8d05da87`, ~814 commits back; 12 `post_model_sync` patches then vs ~78 on the current tip). Most of what follows is about the delivery mechanism, not the intent — and one thing needs deciding before any of it: **#660 and #1276/#1275 solve the same problem by incompatible means**, see the cross-PR section.

---

## What this PR does

It attacks a real, documented defect: `jewellery_erpnext/custom_fields/*.json` is inert because `after_migrate` is commented out at `hooks.py:12`, so the only provisioning path was a hand-maintained 10-record `fixtures/custom_field.json` that had drifted roughly 2,100 fields behind the live sites. The PR (a) replaces the 9-name Custom Field fixture allowlist in `hooks.py` with a module-based filter, (b) deletes `jewellery_erpnext/fixtures/custom_field.json`, and (c) adds a 2,134-record whole-site dump at `jewellery_erpnext/jewellery_erpnext/custom_fields/custom_field.json` (125,908 lines, 179 DocTypes) plus a 16-line `patches/create_custom_fields.py` registered at `jewellery_erpnext/patches.txt:16`.

Several things here are right and worth keeping. The diagnosis is correct. Choosing a registered patch over re-enabling the disabled hook is the right *shape* of fix. `frappe.get_app_path("jewellery_erpnext", "jewellery_erpnext", "custom_fields", "custom_field.json")` resolves correctly (app pymodule dir + module dir). A module filter really is more maintainable than nine hardcoded record names. And committing the field definitions to the repo gives the team a reviewable schema baseline it has never had.

```mermaid
flowchart TD
  A["PR 660"] --> B["hooks.py:292\nfixtures list replaced by one\nCustom Field entry filtered by module"]
  A --> C["delete fixtures/custom_field.json\n10 records"]
  A --> D["add custom_fields/custom_field.json\n2134 records over 179 DocTypes"]
  A --> E["add patches/create_custom_fields.py\nimport_doc loop, no grouping"]
  A --> F["patches.txt line 16\nregisters the patch"]
  D --> E
  F --> G["runs only where jewellery_erpnext\nis already installed"]
  E --> G
```

## Before → after

```mermaid
flowchart LR
  subgraph Before
    B1["bench install-app"] --> B2["sync_fixtures imports\nfixtures/custom_field.json"]
    B3["bench migrate"] --> B2
    B2 --> B4["10 Custom Fields created or refreshed"]
  end
  subgraph After
    A1["bench install-app"] --> A2["set_all_patches_as_completed\nwrites the Patch Log row"]
    A2 --> A3["no fixtures/custom_field.json left"]
    A3 --> A4["0 Custom Fields from this app"]
    A5["bench migrate on an existing site"] --> A6["patch runs import_doc 2134 times\ndelete then re-insert each row"]
  end
```

## Issues at a glance

| 1 | **BLOCKER** | Fresh installs get zero fields: `install-app` pre-marks the patch done and the fixture that did work is deleted | `jewellery_erpnext/patches.txt:16` | every new site, CI site, DR rebuild |
| 2 | **BLOCKER** | `import_doc` skips `CustomField.validate`, so duplicate `(dt, fieldname)` rows land beside gke_customization's | `jewellery_erpnext/patches/create_custom_fields.py:16` | first migrate on any site with both apps |
| 3 | **BLOCKER** | One record whose DocType is absent raises a bare `Exception` and aborts the whole migrate | `jewellery_erpnext/patches/create_custom_fields.py:15` | sites without `payments` — alfarsi, kg-gk |
| 4 | **HIGH** | `import_doc` deletes and re-inserts, so a 3.5-month-old snapshot overwrites site customisation | `jewellery_erpnext/patches/create_custom_fields.py:16` | first migrate on every existing site |
| 5 | **HIGH** | New fixture filter claims three modules owned by sibling apps | `jewellery_erpnext/hooks.py:292` | first `export-fixtures` after merge |
| 6 | **HIGH** | 2,134 per-record `updatedb()` calls, each committing; patch is not resumable | `jewellery_erpnext/patches/create_custom_fields.py:15` | first migrate on every existing site |
| 7 | **MEDIUM** | Patch order differs between fresh and existing sites, so the stale dump can revert later patches | `jewellery_erpnext/patches.txt:16` | on merge, every existing site |
| 8 | **MEDIUM** | Fixtures list is replaced, not amended — Workflow / Role / Stock Entry Type stop being exportable | `jewellery_erpnext/hooks.py:292` | next `export-fixtures` |
| 9 | **MEDIUM** | india_compliance-owned rows are re-stamped `module: "Jewellery Erpnext"` | `jewellery_erpnext/jewellery_erpnext/custom_fields/custom_field.json:40` | first migrate, then on export |
| 10 | **MEDIUM** | `Stock Entry-custom_refining_entry` loses its only provisioning path | `jewellery_erpnext/fixtures/custom_field.json` | fresh provisioning after merge |
| 11 | **MEDIUM** | CI structurally cannot execute either half of this change | `.github/workflows/test.yml:78` | every run on this branch |
| 12 | **LOW** | List-shaped JSON dropped into the directory the `after_migrate` loader reads as dicts | `jewellery_erpnext/jewellery_erpnext/custom_fields/custom_field.json:1` | when `hooks.py:12` is re-enabled |
| 13 | **NIT** | `create_custom_field.py` and `create_custom_fields.py` differ by one character, with opposite implementations | `jewellery_erpnext/patches/create_custom_fields.py:1` | next edit to provisioning |

---

### 1. Fresh installs provision nothing: the patch is pre-marked complete and the fixture is gone · **BLOCKER**

**What** — `bench install-app` writes a Patch Log row for every patch in `patches.txt` *without running it*, so on a brand-new site `jewellery_erpnext.patches.create_custom_fields` is permanently recorded as done and never executes. The PR simultaneously deletes the only mechanism that did run on a fresh install, and adds no `after_install` / `after_sync` entry point.

**Why** — `frappe/installer.py:357-358` calls `set_all_patches_as_completed(name)`, which only inserts Patch Log rows; `frappe/modules/patch_handler.py:56` then filters those out and `:74-76` never reaches the patch again. The fixture path that *did* work is `frappe/installer.py:367 sync_fixtures(name)` → `frappe/utils/fixtures.py:37-45`, which walks the fixtures directory. At this head there is no `after_install` or `after_sync` hook (`git show pr/660:jewellery_erpnext/hooks.py | grep after_` returns only the commented line 12) and no `jewellery_erpnext/install.py`.

**How** — `bench new-site x` → `bench --site x install-app jewellery_erpnext` (Patch Log gains the row, fixtures dir has no `custom_field.json`) → `bench --site x migrate` skips the patch → the app's own provisioning delivers nothing, for good.

**When** — Every new site, CI site, customer onboarding and disaster-recovery rebuild. Existing sites are unaffected, because their Patch Log lacks the row.

**Where** — `jewellery_erpnext/patches.txt:16`, `jewellery_erpnext/patches/create_custom_fields.py:7`

Two honest narrowings, because the headline number is easy to overstate: a fresh site went from **10** provisioned fields to **0** — the 2,134 were never a fresh-install delivery path under either branch. And on a site that also has gke_customization, 1,747 of the dump's 2,134 `(dt, fieldname)` pairs still arrive from gke's own fixture, while `Sketch Order Form-workflow_state` and `Sketch Order-workflow_state` are recreated by `create_custom_field_for_workflow_state` when `fixtures/workflow.json` imports. The measurable gap is the ~387 pairs unique to the dump plus `Stock Entry-custom_refining_entry` (issue 10) — still a failure of this PR's stated purpose.

```python
# frappe/installer.py:353-367
	add_to_installed_apps(name)

	frappe.get_doc("Portal Settings", "Portal Settings").sync_menu()

	if set_as_patched:
		set_all_patches_as_completed(name)

	for after_install in app_hooks.after_install or []:
		frappe.get_attr(after_install)()
...
	sync_jobs()
	sync_fixtures(name)
```

```python
# frappe/installer.py:567-573
def set_all_patches_as_completed(app):
	from frappe.modules.patch_handler import get_patches_from_app

	patches = get_patches_from_app(app)
	for patch in patches:
		frappe.get_doc({"doctype": "Patch Log", "patch": patch}).insert(ignore_permissions=True)
	frappe.db.commit()
```

```mermaid
flowchart TD
  I1["bench install-app jewellery_erpnext"] --> I2["set_all_patches_as_completed\ninstaller.py:358"]
  I2 --> I3["Patch Log now holds\ncreate_custom_fields"]
  I2 --> I4["after_install hooks\nnone declared in this app"]
  I4 --> I5["sync_fixtures\ninstaller.py:367"]
  I5 --> I6["fixtures dir has no custom_field.json"]
  I3 --> I7["every later migrate skips the patch\npatch_handler.py:74-76"]
  I6 --> I8["site has none of this app's fields"]
  I7 --> I8
```

**Fix** — Factor the loop into one idempotent function and register it in **both** places:

```python
# jewellery_erpnext/patches/create_custom_fields.py
def execute():
    from jewellery_erpnext.setup.custom_fields import sync_custom_fields
    sync_custom_fields()
```

```python
# hooks.py
after_install = "jewellery_erpnext.setup.custom_fields.sync_custom_fields"
after_sync = "jewellery_erpnext.setup.custom_fields.sync_custom_fields"
```

`after_install` fires at `frappe/installer.py:360-361`, i.e. *after* `set_all_patches_as_completed`, which makes it the only reliable fresh-site path. This is exactly the shape #1276/#1275 use with their `install.py` — see the cross-PR section before you build it twice.

---

### 2. `import_doc` skips `CustomField.validate`, so duplicate fieldnames land beside gke_customization's rows · **BLOCKER**

**What** — `import_doc` sets `doc.flags.ignore_validate = True`, which makes `Document.run_before_save_methods` return before `CustomField.validate`. That validate is the only thing enforcing "A field with the name X already exists in Y" and `check_fieldname_conflicts`. Where the dump's record *name* differs from the site's existing record name for the same `(dt, fieldname)`, `frappe.db.exists` is False, the existing row is not replaced — it is **joined**, and the DocType ends up with two fields sharing one fieldname.

**Why** — `frappe/modules/import_file.py:232-236` sets the ignore flags; `frappe/model/document.py:1407-1408` returns early on `ignore_validate`, so the duplicate throw at `frappe/custom/doctype/custom_field/custom_field.py:179-184` and `check_fieldname_conflicts(self)` at `:210` never run. `frappe.flags.in_import` (set at `import_file.py:211`) also stops `CustomField.autoname` re-deriving the name, so the dump's name is kept verbatim.

**How** — The dump defines `Warehouse-department` with `fieldname: "department"`; gke_customization's fixture defines `Warehouse-custom_department` with the same fieldname. After the patch, `tabWarehouse` meta carries `department` twice.

**When** — First migrate after merge on any site carrying both apps. The original count of 21 pairs does not survive measurement against live data: most of the 21 share the dump's record name and are therefore replaced rather than duplicated. Measured against `tabCustom Field`, the pairs that actually produce a **second** row are `Warehouse`/`department`, `Sales Invoice`/`sales_invoice`, `Sales Invoice Item`/`customer`, `Sketch Order`/`customer` on alfarsi and kg-gk (kg-gk also `Warehouse`/`subcontractor`), and only `Sketch Order`/`customer` plus `Sales Invoice Item`/`customer` on gk. `Purchase Order`/`purchase_type`, `Stock Entry`/`manufacturer`, `BOM`/`charm`, `Timesheet`/`order` and the rest do **not** fire.

**Where** — `jewellery_erpnext/patches/create_custom_fields.py:16`

```python
# frappe/modules/import_file.py:229-238
	if frappe.db.exists(doc.doctype, doc.name):
		delete_old_doc(doc, reset_permissions)

	doc.flags.ignore_links = True
	if not data_import:
		doc.flags.ignore_validate = True
		doc.flags.ignore_permissions = True
		doc.flags.ignore_mandatory = True

	doc.insert()
```

```python
# frappe/custom/doctype/custom_field/custom_field.py:179-184
			if self.is_new() and self.fieldname in fieldnames:
				frappe.throw(
					_("A field with the name {0} already exists in {1}").format(
						frappe.bold(self.fieldname), self.dt
					)
				)
```

One verification pass took this further, and I would like your read on it before we grade the blast radius. Fixture imports run **with** validation (`frappe/utils/fixtures.py:45` → `data_import.import_doc` → `import_file_by_path(f, data_import=True, ...)`, so `import_file.py:232 if not data_import:` is False), and `sync_fixtures()` runs at `frappe/migrate.py:169-171` — later in the **same** migrate as the post_model_sync patches at `frappe/migrate.py:136-145`. **Can you confirm that on a site carrying both apps, the duplicate this patch creates makes gke's re-import of `Warehouse-custom_department` throw `ValidationError` at "Syncing fixtures...", and that `fixtures.py:46` catching only `(ImportError, frappe.DoesNotExistError)` lets it abort the migrate?** If so, the failure is not a deferred one at the next Custom Field save — it is the same migrate, with the patch's rows already committed.

```mermaid
flowchart TD
  M1["bench migrate"] --> M2["run_schema_updates\npost_model_sync patches\nmigrate.py:136-145"]
  M2 --> M3["patch inserts Warehouse-department\nvalidation skipped"]
  M3 --> M4["tabWarehouse has two rows\nwith fieldname department"]
  M4 --> M5["post_schema_updates\nsync_fixtures\nmigrate.py:169-171"]
  M5 --> M6["gke fixture re-imports\nWarehouse-custom_department\nWITH validation"]
  M6 --> M7["CustomField.validate throws"]
  M7 --> M8["fixtures.py:46 catches only\nImportError and DoesNotExistError"]
  M8 --> M9["migrate aborts, duplicate row persists"]
```

**Fix** — Do not import Custom Fields with `import_doc`. Reshape the dump to `{doctype: [field dicts]}` and call `frappe.custom.doctype.custom_field.custom_field.create_custom_fields(...)`, which keys on `(dt, fieldname)` via `get_existing_custom_fields` and therefore cannot create a second row for one fieldname. Separately, resolve the name collisions with gke_customization explicitly: decide which app owns each field and drop it from the other.

---

### 3. A record whose parent DocType is absent raises a bare `Exception` and aborts the migrate · **BLOCKER**

**What** — The loop has no `try`/`except` and no existence check on `doc["dt"]`. `frappe.db.updatedb` raises a plain `Exception` when the DocType row is missing, so one unresolvable record kills `bench migrate` for the whole bench.

**Why** — `frappe/database/mariadb/database.py:468-470`. Because `import_doc` sets `ignore_links` and `ignore_validate`, the insert itself succeeds and the failure surfaces in `on_update`. Frappe's own fixture importer guards exactly this case at `frappe/utils/fixtures.py:46-48`; this patch does not.

**When** — The broad version of this claim did not survive verification: I checked all 179 distinct `dt` values in the dump against `tabDocType` on alfarsi, gk and kg-gk. `lending` is installed everywhere, so none of the `Loan*` records fire, and `SI Item` exists on all three (it ships in this app's own `fixtures/doctype.json`). Exactly one record is left: **`GoCardless Mandate`**, from the `payments` app, which alfarsi and kg-gk do not have. **Can you confirm that `bench --site alfarsi migrate` and `bench --site kg-gk migrate` abort on that record?** If they do, this is a failed migration on a site that transacts daily, and note CI cannot catch it — `install.sh:88` installs `payments` on `test_site`.

**Where** — `jewellery_erpnext/patches/create_custom_fields.py:15`

```python
# jewellery_erpnext/patches/create_custom_fields.py:12-16 (pr/660)
	with open(path) as f:
		docs = json.load(f)

	for doc in docs:
		import_doc(doc, ignore_version=True)
```

```python
# frappe/database/mariadb/database.py:461-470
	def updatedb(self, doctype, meta=None):
		"""
		Syncs a `DocType` to the table
		...
		"""
		res = self.sql("select issingle from `tabDocType` where name=%s", (doctype,))
		if not res:
			raise Exception(f"Wrong doctype {doctype} in updatedb")
```

```python
# the framework's own guard, which this patch omits — frappe/utils/fixtures.py:44-48
		try:
			import_doc(file_path, sort=True)
		except (ImportError, frappe.DoesNotExistError) as e:
			# fixture syncing for missing doctypes
			print(f"Skipping fixture syncing from the file {fname}. Reason: {e}")
```

**Fix** — Skip records whose parent DocType is absent, and never let one record abort a migrate:

```python
for doc in docs:
    if not frappe.db.exists("DocType", doc["dt"]):
        frappe.log_error(f"create_custom_fields: skipping {doc['name']}, DocType {doc['dt']} not installed")
        continue
    try:
        import_doc(doc, ignore_version=True)
    except Exception:
        frappe.log_error(title=f"create_custom_fields failed for {doc['name']}")
```

Independently, strip the `GoCardless Mandate` and `Loan*` records from the dump — they are not this app's to ship.

---

### 4. `import_doc` deletes and re-inserts, so a 3.5-month-old snapshot overwrites site customisation · **HIGH**

**What** — `import_doc` is not an upsert. Where the record name already exists it calls `delete_old_doc()` → `frappe.delete_doc(..., force=1, for_reload=True)` and then re-inserts the dumped values wholesale, with validation, permission and mandatory checks off. Replace semantics, not merge.

**Why** — `frappe/modules/import_file.py:229-238` and `:272-273`. `for_reload=True` skips permissions and `on_trash` and forces `delete_permanently`, so there is not even a Deleted Document snapshot to recover from. The framework's own helper merges field-by-field instead (`frappe/custom/doctype/custom_field/custom_field.py:322`, `custom_field.update(df)`).

**How** — Measured by diffing all 2,134 dump records against live `tabCustom Field` over label / fieldtype / insert_after / options / depends_on / reqd / read_only / hidden / in_list_view / precision / length / permlevel / module: **alfarsi 1,100 of 2,044 existing rows differ, gk 1,122 of 2,126, kg-gk 1,126 of 2,038.** Top differing columns on gk: `module` 873, `options` 134, `insert_after` 119, `hidden` 54, `depends_on` 52, `precision` 36, `length` 27, `label` 22. Concrete reverts on gk: `BOM-diamond_ratio` precision 4 → blank, `Stock Entry-customer` hidden 0 → 1, `Sales Order Item-bom` read_only 0 → 1.

**When** — First migrate after merge on every existing site, including kg-gk, which is live.

Scope narrowing worth stating: 1,627 of the 2,134 records are `is_system_generated=1`, and Customize Form does not write those Custom Field rows at all — it writes Property Setters (`customize_form.py:501` short-circuits on `is_standard_or_system_generated_field`), and Property Setters survive this patch because `delete_doc(for_reload=True)` skips `CustomField.on_trash`. The irreversible loss is therefore bounded to the **507 non-system-generated records** plus anything edited directly on the Custom Field form. Real and serious; not 2,134-wide.

**Where** — `jewellery_erpnext/patches/create_custom_fields.py:16`

```python
# frappe/modules/import_file.py:272-273
	# delete old
	frappe.delete_doc(doc.doctype, doc.name, force=1, ignore_doctypes=ignore, for_reload=True)
```

```mermaid
flowchart LR
  R1["dump record"] --> R2{"frappe.db.exists\nCustom Field name"}
  R2 -- "yes" --> R3["delete_old_doc\nforce=1 for_reload=True"]
  R3 --> R4["insert the dump's values\nvalidation and permissions off"]
  R2 -- "no" --> R4
  R4 --> R5["site edits on the 507 non\nsystem generated rows are gone"]
  R6["create_custom_fields update=False"] --> R7["keyed on dt and fieldname\nexisting row untouched"]
```

**Fix** — `create_custom_fields(fields_by_doctype, update=False)`, so existing rows are left alone and only missing fields are created — that is what "provisioning creates, never modifies" means here. If specific fields genuinely must be corrected, list those in a separate, reviewable patch rather than rewriting everything.

---

### 5. The new fixture filter claims three modules this app does not own · **HIGH**

**What** — The filter widens from nine hardcoded names to `module in ["Jewellery Erpnext", "Gurukrupa Exports", "GKE Order Forms", "Gurukrupa Customizations", "GKE Catalog"]`. `GKE Order Forms` and `GKE Catalog` are exactly the two modules `gke_customization/gke_customization/hooks.py:319-322` already exports as its own Custom Field fixture; `Gurukrupa Customizations` belongs to a third app. Module census of the dump: `Jewellery Erpnext` 1,958, **`Gurukrupa Customizations` 91, `GKE Order Forms` 70, `GKE Catalog` 3**, `Gurukrupa Exports` 12 — **164 sibling-owned records**.

**Why** — Export iterates the hook (`frappe/utils/fixtures.py:67-79`); import iterates the *directory* (`:37-45`). So the next `bench export-fixtures --app jewellery_erpnext` writes a ~2,100-record `fixtures/custom_field.json` back into this app, and from then on every migrate imports both files, overlapping on **1,747 `(dt, fieldname)` pairs**. Last app in `sites/apps.txt` wins, which is not a contract. The base branch declined to do this on purpose — the module-filter variant exists at `upstream/v16_develop_aerele:jewellery_erpnext/hooks.py:481-483` and is commented out.

**When** — Nothing ships the moment this merges, because the PR deletes the fixture file. The war starts with the first `bench export-fixtures` after merge, on every site with both apps. The exposure is then larger than 164: on gk the patch re-stamps ~873 more rows into `module: "Jewellery Erpnext"` (294 from `GST India`, 252 from `GKE Order Forms`, 327 from NULL), and those would match the new filter too.

**Where** — `jewellery_erpnext/hooks.py:292`

```python
# jewellery_erpnext/hooks.py:292-308 (pr/660)
fixtures = [
	{
		"doctype": "Custom Field",
		"filters": {
			"module": [
				"in",
				[
					"Jewellery Erpnext",
					"Gurukrupa Exports",
					"GKE Order Forms",
					"Gurukrupa Customizations",
					"GKE Catalog",
				],
			]
		},
	}
]
```

```python
# apps/gke_customization/gke_customization/hooks.py:319-322
fixtures = [
    {
        "dt": "Custom Field",
        "filters": [["module", "in", ["GKE Order Forms", "GKE Catalog"]]],
    },
```

```mermaid
flowchart LR
  X1["jewellery hooks.py:292\nfive modules claimed"] --> X2["bench export-fixtures writes\njewellery fixtures/custom_field.json"]
  X3["gke hooks.py:319-322\nGKE Order Forms and GKE Catalog"] --> X4["gke fixtures/custom_field.json\n2082 records"]
  X2 --> X5["import_fixtures walks the directory\non every migrate"]
  X4 --> X5
  X5 --> X6["1747 shared dt and fieldname pairs\nlast app in apps.txt wins"]
```

**Fix** — Narrow the filter to `["Jewellery Erpnext", "Gurukrupa Exports"]`, and filter the 164 sibling-owned records out of the dump before committing it. If jewellery genuinely needs one of those fields, it belongs in the owning app.

---

### 6. 2,134 per-record `updatedb()` calls, each committing, and the patch is not resumable · **HIGH**

**What** — The loop never sets `frappe.flags.in_create_custom_fields`, so `CustomField.on_update` runs `frappe.clear_cache(doctype=self.dt)` and `frappe.db.updatedb(self.dt)` for every one of the 2,134 records — and `MariaDBDatabase.updatedb` ends with `self.commit()`. The framework's bulk helper sets that flag at `custom_field.py:336` precisely so the schema is synced once per doctype: **179 syncs instead of 2,134.**

**Why** — `frappe/model/document.py:1452-1453` runs `on_update` for `_action == "save"` regardless of `ignore_validate` (that only short-circuits `run_before_save_methods`), so `custom_field.py:212-222` executes every iteration. Doctype census: BOM 268, Item 216, Stock Entry 113, Sales Order Item 91, Employee 91, Quotation Item 90, Customer 84, Sales Invoice 73, Company 62, Stock Entry Detail 61.

**How** — Two measurable consequences, stated conservatively. (i) 2,134 cache flushes, schema-diff round trips and COMMITs in one migrate, which is the single longest step in it. (ii) **Non-resumability**: the Patch Log row is only written after the whole loop, while the per-record `commit()` has already persisted records 1..N-1 — so a failure at record N (see issue 3) replays the entire file as delete-and-reinsert on the next attempt. I would not lean on the atomicity framing: MariaDB commits DDL implicitly, so `@atomic` on `run_schema_updates` is already nominal for any field-creating patch. And since `for_reload=True` skips `on_trash`, `MariaDBTable.sync()` finds no column diff on sites that already have the fields — so this is a one-off long window, not repeated `tabBOM`/`tabItem` rebuilds.

**When** — Every existing site, once.

**Where** — `jewellery_erpnext/patches/create_custom_fields.py:15`

```python
# frappe/custom/doctype/custom_field/custom_field.py:212-222
	def on_update(self):
		# validate field
		if not self.flags.ignore_validate:
			from frappe.core.doctype.doctype.doctype import validate_fields_for_doctype

			validate_fields_for_doctype(self.dt)

		# clear cache and update the schema
		if not frappe.flags.in_create_custom_fields:
			frappe.clear_cache(doctype=self.dt)
			frappe.db.updatedb(self.dt)
```

```python
# frappe/custom/doctype/custom_field/custom_field.py:383-388 — the bulk path, one sync per doctype
		for doctype in doctypes_to_update:
			frappe.clear_cache(doctype=doctype)
			frappe.db.updatedb(doctype)

	finally:
		frappe.flags.in_create_custom_fields = False
```

**Fix** — Group by `dt` and call `create_custom_fields(fields_by_dt, update=False)`; it sets the flag, batches per doctype and restores `(dt, fieldname)` upsert semantics, which also closes issues 2 and 4. If a manual loop must stay, wrap it in `frappe.flags.in_create_custom_fields = True` / `finally` and do one `updatedb` per distinct `dt` at the end.

---

### 7. Patch order differs between fresh and existing sites, so the stale dump can revert later patches · **MEDIUM**

**What** — The entry is inserted at position 12 of the branch's `post_model_sync` list, before `backfill_employee_ir_mop_log_to_warehouse`, not appended. On merge onto the current tip it would sit ahead of roughly 62 upstream patches, many of which create or alter specific custom fields. But `patch_handler` skips anything already in Patch Log, so on an **existing** site — where those 62 have run — this one executes **last**, and a delete-and-reinsert of a 3.5-month-old snapshot is then authoritative.

**Why** — `frappe/modules/patch_handler.py:56` and `:74-76`: file position only orders patches that are all pending. Combined with `import_file.py:229-238`, running last means the snapshot wins.

**How** — Verified reverts on gk: `Sales Order Item-custom_gross_weight` currently has `insert_after: "column_break_ihdh"` (a layout-patch value) while the dump has `purchase_order_item`; `Quotation-finding_weight_for_chain` is `custom_total_rows` on gk vs `items` in the dump. `compact_sales_weight_field_layout` is the clearest instance of a later patch being undone. (The `widen_stock_entry_jwelex_tag_field` case does *not* apply — the dump has no `Stock Entry Detail-custom_jwelex_tag_no` record.)

**When** — On merge, for every existing site; the fresh-vs-existing divergence is permanent. Note `hooks.py` (+212) and `patches.txt` (+70) both moved upstream since the merge base, so this arrives as a merge conflict a human resolves rather than silently.

**Where** — `jewellery_erpnext/patches.txt:16`

```mermaid
flowchart TD
  P1["bench migrate after merge"] --> P2{"has this site already run\nthe later upstream patches"}
  P2 -- "existing site" --> P3["only create_custom_fields is pending"]
  P3 --> P4["snapshot runs LAST and overwrites\nwhat those patches set"]
  P2 -- "fresh site" --> P5["all patches pending, file order applies"]
  P5 --> P6["snapshot runs 12th, later patches\ncorrect it afterwards"]
  P4 --> P7["two sites, different field definitions,\nno diff to explain it"]
  P6 --> P7
```

**Fix** — Rebase onto current `v16_develop_aerele`, regenerate the dump from a site that has run every patch, and **append** the entry rather than inserting it at 12. Then make the patch non-destructive (`update=False`), which is the only fix that survives the next stale rebase — ordering stops being load-bearing at all.

---

### 8. The fixtures list is replaced, not amended — Workflow, Role and Stock Entry Type stop being exportable · **MEDIUM**

**What** — `fixtures` becomes a single Custom Field entry. The `Workflow` (2 named workflows), `Workflow State`, `Workflow Action Master` and `Role` (`GK sales user`, `Sketch QC`, `All`) declarations are dropped, while `fixtures/workflow.json`, `workflow_state.json`, `workflow_action_master.json` and `role.json` all remain on disk. On merge onto the current tip the same replacement also removes `Refining Entry Workflow`, the `Refining User` / `Refining Manager` roles, and the `"Stock Entry Type"` entry that the base added with a six-line comment explaining exactly why it must stay exportable.

**Why** — Import is directory-driven (`frappe/utils/fixtures.py:37-45`), export is hook-driven (`:67-79`). Removing a hook entry removes only the export half: the files keep being delete-and-reinserted on every migrate, with no supported way to regenerate them.

**How** — Someone adds a workflow state or a role permission on a configured site and runs `bench export-fixtures`. Nothing is written. The next migrate re-imports the stale on-disk copy and reverts the change, with no diff in the repo to explain it.

**When** — First `export-fixtures` after merge.

**Where** — `jewellery_erpnext/hooks.py:292`

```python
# upstream/v16_develop_aerele:jewellery_erpnext/hooks.py:474-480 — the comment this replacement removes
	# fixtures/stock_entry_type.json is imported on every migrate regardless of this
	# list (import_fixtures walks the whole directory), and that import deletes and
	# re-inserts each record -- so custom_allowed_roles set in the desk on a record
	# present in the file is wiped unless the file carries it. Listing it here makes
	# the file re-exportable, so `bench export-fixtures` can regenerate it from a
	# configured site instead of it being hand-edited.
	"Stock Entry Type",
```

**Fix** — Keep every existing entry and change only the Custom Field one. On top of current `v16_develop_aerele` that means leaving the `Workflow` / `Workflow State` / `Workflow Action Master` / `Role` entries and `"Stock Entry Type"` untouched, replacing only the 10-name Custom Field block, and narrowing its module list per issue 5.

---

### 9. india_compliance-owned rows are re-stamped `module: "Jewellery Erpnext"` · **MEDIUM**

**What** — The dump carries GST fields that india_compliance creates and owns, stamped `"module": "Jewellery Erpnext"` and `"is_system_generated": 1`. Measured against live data on gk, the patch rewrites **294 rows currently carrying `module = "GST India"`**, plus 252 currently `GKE Order Forms` and 327 currently NULL, into `Jewellery Erpnext`.

**Why** — `module` is the ownership marker. Rewriting it in an export does not transfer ownership, it hides the conflict — and the re-stamped rows then match this PR's new module filter (issue 5), so a later `export-fixtures` pulls another app's fields into jewellery's fixture permanently.

Two claims from the original write-up do not hold, and I would rather not overstate this: `Purchase Invoice-gst_section` already has `insert_after: "language"` on alfarsi, gk and kg-gk, so importing the dump moves nothing; and there is no ping-pong, because `india_compliance/hooks.py:31` sets `after_migrate` to the audit-trail setup, which does not re-create custom fields. The defensible harm is the one-way ownership re-stamp.

**When** — First migrate after merge on every site with india_compliance; the export consequence follows the first `export-fixtures`.

**Where** — `jewellery_erpnext/jewellery_erpnext/custom_fields/custom_field.json:40`

```python
# apps/india_compliance/india_compliance/gst_india/constants/custom_fields.py:995-1000
    "Purchase Invoice": [
        {
            "fieldname": "gst_section",
            "label": "GST Details",
            "fieldtype": "Section Break",
            "insert_after": "gst_vehicle_type",
```

**Fix** — Strip every record whose fieldname is declared in `india_compliance/gst_india/constants/custom_fields.py`, and more generally every `is_system_generated: 1` record this app's own code did not create, out of the dump before shipping it.

---

### 10. `Stock Entry-custom_refining_entry` loses its only provisioning path · **MEDIUM**

**What** — The deleted fixture held 10 records. Nine reappear in the new dump; `Stock Entry-custom_refining_entry` does not, and it is absent from gke_customization's fixture too. No patch on `upstream/v16_develop_aerele` creates it.

**Why** — Its `module` is NULL, so even after a re-export the new module-based filter could never pick it up. Meanwhile the upstream Refining module still uses it: `refining_entry.json:632` has `"link_fieldname": "custom_refining_entry"`, and `refining_entry.py` writes it on the Stock Entries it creates.

**How** — Merge onto current `v16_develop_aerele` and provision a fresh site: the field is never created, and the Refining Entry dashboard's Stock Entry link resolves against a column that does not exist.

**When** — Fresh provisioning after merge. Existing sites keep the field. The Refining module does not exist at this branch's tree — `git grep custom_refining_entry pr/660` returns nothing — which is exactly why this is easy to miss. Note that deleting the fixture will surface as a modify/delete conflict on merge (upstream added 77 lines to that file since the merge base), so the outcome only lands if the resolver keeps the deletion.

**Where** — `jewellery_erpnext/fixtures/custom_field.json` (deleted)

**Fix** — Regenerate the dump from a site that has the Refining module, or add the record by hand before deleting the fixture. Worth adding a merge gate: assert that every `name` in the deleted fixture is present in the new dump. That one check catches this class of loss in a 125k-line diff.

---

### 11. CI structurally cannot execute either half of this change · **MEDIUM**

**What** — The PR adds no `test_*.py` and edits no workflow, and the pipeline could not exercise it even if it wanted to. `install.sh:95` runs `bench --site test_site install-app jewellery_erpnext`, which marks `create_custom_fields` complete, so the `bench --site test_site migrate` at `.github/workflows/test.yml:78` skips it — **the 125,908-line dump is never imported in CI.** A green check on this PR is evidence of nothing.

**Why** — Same mechanism as issue 1: `frappe/installer.py:358` → `set_all_patches_as_completed`, then `frappe/modules/patch_handler.py:74-76` skips the logged patch.

**When** — Every run on this branch and on anything that merges it.

**Where** — `.github/workflows/test.yml:78`, `install.sh:95`

There is a second half I could not verify from this bench, so treat it as a question: `install.sh:79-81` does an unconditional `cp -rf` of `fixture_source/jewellery_erpnext/fixtures/*` over the app's fixtures directory, from the external `git_action_v16` checkout. **If that branch still carries `custom_field.json`, CI also never sees the fixture deletion.** Can you check?

```bash
# install.sh:79-81 (pr/660)
cp -rf \
"$GITHUB_WORKSPACE/fixture_source/jewellery_erpnext/fixtures/"* \
"$HOME/frappe-bench/apps/jewellery_erpnext/jewellery_erpnext/fixtures/"
```

**Fix** — Add a regression test and register it in `.github/workflows/test.yml` in this same PR. Minimum useful set: every `dt` in the dump exists as a DocType; every `module` in the dump is one this app owns; running the patch twice is idempotent and creates a known field. Do not override `setUpClass` with a bare `pass` — that class writes to the DB, and the bare override disables the class-scoped rollback. Also add an explicit step that actually runs the patch after migrate (`bench --site test_site execute jewellery_erpnext.patches.create_custom_fields.execute`), otherwise the dump stays untested whatever else changes.

---

### 12. List-shaped JSON dropped into the directory the `after_migrate` loader reads as dicts · **LOW**

**What** — `jewellery_erpnext/migrate.py:23-32` iterates every file in `jewellery_erpnext/custom_fields/` and does `CUSTOM_FIELDS.update(json.load(f))`, i.e. it requires each file to be a `{doctype: [field dicts]}` mapping — which the 41 existing files are. The new `custom_field.json` in that same directory is a JSON **list** of 2,134 flat records, so `dict.update` would raise `ValueError: dictionary update sequence element #0 has length 57; 2 is required`.

**When** — Latent today: `after_migrate` is commented out at `hooks.py:12`, and the sibling patch `patches/create_custom_field.py` hard-filters to three named files, so nothing hits it on this branch. It fires the moment someone re-enables `hooks.py:12` — which is exactly the direction this work is heading.

**Where** — `jewellery_erpnext/jewellery_erpnext/custom_fields/custom_field.json:1`

```python
# git show pr/660:jewellery_erpnext/migrate.py  (lines 23-32)
def create_custom_fields():
	CUSTOM_FIELDS = {}
	print("Creating/Updating Custom Fields....")
	path = os.path.join(os.path.dirname(__file__), "jewellery_erpnext/custom_fields")
	for file in os.listdir(path):
		with open(os.path.join(path, file), "r") as f:
			CUSTOM_FIELDS.update(json.load(f))
	from frappe.custom.doctype.custom_field.custom_field import create_custom_fields

	create_custom_fields(CUSTOM_FIELDS)
```

**Fix** — Reshaping the dump to `{doctype: [field dicts]}` fixes this and issues 2, 4 and 6 at the same time. If the flat list has to stay, put it where that directory loader does not scan it.

---

### 13. Two patch modules one character apart, with opposite implementations · **NIT**

`jewellery_erpnext/patches/create_custom_field.py` (singular) already exists — it reads three named files from `custom_fields/` and calls frappe's `create_custom_fields()` helper. This PR adds `create_custom_fields.py` (plural), which reads a different file from the same directory and uses `import_doc`. Only the plural one is registered, at `jewellery_erpnext/patches.txt:16`, and the unregistered singular one is the version using the correct helper. Give the new patch a name that says what it does and when it ran (e.g. `patches/v16_import_custom_field_snapshot.py`), and delete the singular one if it is dead.

---

## Tests & CI

No `test_*.py` is added or changed by this PR, and no workflow file is touched — the test diff is zero bytes, and `.github/workflows/test.yml` is the only workflow at this head. Beyond "no tests", the pipeline cannot reach the change at all: `install.sh:95` pre-marks the patch complete and `test.yml:78` is the only migrate, so the dump is never imported in CI (issue 11). Nothing here is caught by a green check — not the duplicate-fieldname collision, not the missing-DocType abort, not the 2,134-row rewrite.

The minimum to make this reviewable in CI is in issue 11: a dump-validity test module plus an explicit step that executes the patch against `test_site`, both registered in `test.yml` in this PR.

**#660 vs #1276 / #1275 — architecturally incompatible; this is the decision to make first.** All three fix "custom_fields JSON never reaches a site", by opposite means. #1276 (`kggk_uat`) and #1275 (`kggk_prod`) add a 677-line `install.py` wired to `after_install` + `after_sync`, with explicit logic to *defer* fields a sibling app's fixtures own. #660 deletes `fixtures/custom_field.json` and ships a whole-site dump plus a patch that imports all of it, sibling-app records included. **Whichever merges second silently breaks the other**: the dump re-asserts the very fields `install.py` deliberately defers, and `install.py` re-asserts them back on the next `after_sync`. Note also that `install.py` does not exist at this PR's base — the fix suggested in issue 1 is the same shape those PRs already implement, so it should be built once, not twice.

Contention on shared files: `jewellery_erpnext/patches.txt` is touched by **12** open PRs and `jewellery_erpnext/hooks.py` by **10**, both including #660. `patches.txt` is the worst merge-order hazard in the repo — everyone appends to the same tail, and order is load-bearing here (issue 7). #660 makes it worse than most by *inserting* mid-list rather than appending.

```mermaid
flowchart LR
  C1["PR 660\n2134-record dump plus patch\nfixture deleted"] --> C3{"same defect:\ncustom_fields JSON\nnever reaches a site"}
  C2["PR 1276 and PR 1275\ninstall.py on after_install and after_sync\ndefers sibling-owned fields"] --> C3
  C3 --> C4["whichever merges second\nsilently breaks the other"]
  C4 --> C5["decide the architecture\nbefore either lands"]
```

## Suggested order of work

1. **Settle the architecture against #1276/#1275** before spending more on this branch. If `install.py` is the agreed mechanism, this PR's useful residue is the committed field definitions, not the patch.
2. **Rebase onto current `v16_develop_aerele`** and regenerate the dump from a site that has run every patch. That also forces the `hooks.py` and `patches.txt` conflicts into the open rather than leaving them for the merge.
3. **Filter the dump**: drop the 164 sibling-app records, the india_compliance-owned rows, `GoCardless Mandate` and the `Loan*` records. Add back `Stock Entry-custom_refining_entry`, or state in the description that it is dead.
4. **Reshape to `{doctype: [field dicts]}` and switch to `create_custom_fields(..., update=False)`** — one change that closes issues 2, 4, 6 and 12.
5. **Register the provisioning in both places** — `patches.txt` (appended, not inserted) and an `after_install`/`after_sync` hook — so fresh sites are covered.
6. **Narrow the fixtures filter** to `["Jewellery Erpnext", "Gurukrupa Exports"]` and restore the `Workflow` / `Workflow State` / `Workflow Action Master` / `Role` / `Stock Entry Type` entries.
7. **Add the dump-validity tests and the workflow registration**, plus a CI step that actually runs the patch.

Happy to go deeper on any of these — particularly the fixture-sync abort in issue 2 and the `GoCardless Mandate` case in issue 3, where a quick migrate on a scratch clone of alfarsi would settle both.

PR #1043, opened 2026-08-04
Feat/job work order

Kanishkar16Vijay:


DHINESH00:
## Review — #1043 Feat/job work order

**Verdict:** Needs rework
&nbsp;·&nbsp; 3 blocker · 3 high · 9 medium · 2 low · 1 nit
&nbsp;·&nbsp; base `v16_develop_aerele` · head `3f309867` · 9 files, +415/-20

**Start here — this is the only open PR with a red check.** The failure is in the
`make_quotation` changes at `doc_events/purchase_order.py:361-374`: the new PMO requirement
(`:373-374`) plus the new `po_doc.supplier` dereference (`:362`) both run unconditionally, and
`jewellery_erpnext.jewellery_erpnext.tests.test_purchase_order` is in the CI allowlist
(`.github/workflows/core-tests.yml:65`). All five `make_quotation` tests in that module now fail —
the first one to bite is `po_doc.supplier` at `:362`, which raises `AttributeError` before the
`custom_pmo` throw is ever reached. Details in issue 4.

---

## What this PR does

It wires a job-work / subcontracting lane across six server modules: Manufacturing Plan `on_submit`
auto-creates and submits a mirror "Employee IR" Sales Order; `make_quotation` re-sources the
Quotation from the PO row's PMO and resolves the internal Company from `Company.supplier_code`;
Purchase Invoice gains a `before_validate` hook that back-fills item rates from the matching Sales
Invoice; Employee IR Receive is gated on an existing Purchase Invoice; Stock Entry gains a
"Customer Goods Issue" lane with a new `make_customer_goods_issue` builder on MWO; and SNC gains an
`include_all_items` mode plus a "Create SNC (All Materials)" button.

Things this does well, and they are worth keeping as the rework happens: every new
`frappe.db.sql` is parameterised with `%s` — no interpolation, no injection introduced; the new
whitelisted method is a document method, so `run_doc_method`'s `check_permission=True` covers it;
`create_employee_ir_sales_orders` groups by source Sales Order so one SO is created per SO rather
than one per row, and it carries an explicit idempotency guard instead of hoping; `valid_pos` uses
`employee_ir != ""`, which correctly also excludes NULL in MariaDB; the Customer Goods Issue draft
reuses the existing `flags.throw_batch_error` convention rather than inventing a new one; the new
Quotation mapper fails loudly with actionable messages instead of silently writing blanks; and
sourcing item/qty/branch/project/BOM from the Parent Manufacturing Order really is the right source
of truth for a manufacturing-driven quotation. There is also no new `frappe.db.commit()` anywhere in
the diff.

```mermaid
flowchart TD
  A["Manufacturing Plan.on_submit"] --> B["create_employee_ir_sales_orders\nprocess_so_for_employee_ir"]
  B --> C["new Sales Order, item 'Subcontracting Income'\ninsert plus submit"]
  D["Employee IR type Issue, subcontracting Yes"] --> E["create_subcontracting_order\nPurchase Order carrying custom_pmo"]
  E --> F["PO to Quotation\nmake_quotation now PMO sourced"]
  E --> G["Purchase Invoice before_validate\nrate back filled from the Sales Invoice"]
  H["Employee IR type Receive"] --> I["validate_purchase_invoice gate\nrequires a submitted PI on the Issue IR PO"]
  J["MWO button 'Customer Goods Issue'"] --> K["make_customer_goods_issue\nStock Entry built from MOP Log balances"]
  K --> L["Stock Entry submit\nMWO status set to Completed"]
  M["MWO button 'Create SNC All Materials'"] --> N["create_snc include_all_items\nadmits F-G, D- and G- rows"]
```

## Before → after

The two behaviour changes that reach users who are not in the job-work flow at all:

```mermaid
flowchart LR
  subgraph "Before"
    B1["Create Quotation from any PO"] --> B2["rows mapped from the PO row values"]
    B3["Save any Purchase Invoice"] --> B4["validate hooks only\nrates untouched"]
  end
  subgraph "After"
    A1["Create Quotation from any PO"] --> A2["look up Company by supplier_code\nthen require custom_pmo on every row"]
    A2 --> A3["PO without a PMO: hard throw"]
    A4["Save any Purchase Invoice"] --> A5["before_validate runs\nupdate_expense_account\nupdate_rate_from_sales_invoice"]
    A5 --> A6["row.rate overwritten from a Sales Invoice"]
  end
```

Both new paths are unconditional for their doctype — there is no `purchase_type`, company or
`employee_ir` gate on either.

## Issues at a glance

| 1 | **BLOCKER** | `batch_voucher_type` deleted from the SNC row dict; the Customer Repair settlement exclusion is now dead | `snc.py:579` | Any Create SNC on an MWO holding a Customer Repair batch — default path, no flag |
| 2 | **BLOCKER** | New PI `before_validate` queries `Sales Order Item.custom_design_code`, a column nothing creates | `purchase_invoice.py:26` | Saving any Purchase Invoice with a PO-linked item row |
| 3 | **BLOCKER** | MP submit joins `soi.custom_parent_manufacturing_order`, also never created | `manufacturing_plan.py:177` | First Manufacturing Plan submit on the Employee-IR path |
| 4 | **HIGH** | `make_quotation` requires `custom_pmo` and dereferences `po_doc.supplier` unconditionally — the red CI check | `purchase_order.py:373` | Every CI run; every PO→Quotation for a PO without a PMO |
| 5 | **HIGH** | Customer Goods Issue submit force-writes MWO `status = "Completed"`, no reversal on cancel | `stock_entry.py:935` | Every Customer Goods Issue submit, and every cancel of one |
| 6 | **HIGH** | PI rates rewritten from an arbitrarily-picked Sales Invoice, drafts included | `purchase_invoice.py:42` | Every PI save with a PO-linked row, once issue 2 is fixed |
| 7 | **MEDIUM** | Idempotency guard keys on a nullable link — duplicate submitted Sales Order on amend | `manufacturing_plan.py:219` | Amend or re-submit of an Employee-IR Manufacturing Plan |
| 8 | **MEDIUM** | Auto-created Sales Order can be submitted at rate 0; SSP lookup unordered and unfiltered | `manufacturing_plan.py:230` | No Party Link, no `item_subcategory`, or two SSP rows per supplier |
| 9 | **MEDIUM** | Hardcoded `"Subcontracting Income"` Item that no fixture or patch creates | `manufacturing_plan.py:248` | Any site without that Item — it exists on `gk` only |
| 10 | **MEDIUM** | MOP Log query filters `docstatus` on a non-submittable doctype and ignores `is_cancelled` | `manufacturing_work_order.py:540` | Every use of the new button on an MWO with a cancelled row |
| 11 | **MEDIUM** | Receive gate runs in `validate` and picks the Issue IR with an unordered `LIMIT 1` | `employee_ir.py:177` | Every draft save of a subcontracting Receive IR |
| 12 | **MEDIUM** | Receive may be permanently blocked when the Issue IR produced no Purchase Order | `employee_ir.py:214` | Issue IR whose PO had no item rows |
| 13 | **MEDIUM** | `set_repeat_receive_flag()` re-indented into `validate_purchase_invoice`, skipped by two early returns | `employee_ir.py:220` | Receive IR with an empty or MWO-less operations table |
| 14 | **MEDIUM** | `include_all_items` puts carats into `custom_pure_qty` and then hard-throws on the same rows | `snc.py:579` | "Create SNC (All Materials)" on an MWO with non-gold rows |
| 15 | **MEDIUM** | 415 lines, zero tests, zero workflow registration | repo-wide | Now, and on every later change to these modules |
| 16 | **LOW** | `include_all_items` arrives as a string, so `"0"` enables all-items mode | `snc.py:164` | Any caller other than the new button |
| 17 | **LOW** | Three round trips per row plus a nested Sales Order insert and submit inside `on_submit` | `manufacturing_plan.py:213` | Every Employee-IR plan submit; cost scales with plan size |
| 18 | **NIT** | Loop-body import, leftover debug throw, untranslated messages, stray blank line | several | — |

---

### 1. `batch_voucher_type` dropped from the SNC row dict — Customer Repair gold is now settled  ·  **BLOCKER**

**What** — The line `"batch_voucher_type": batch_voucher_type,` was deleted from the dict
`_get_receivable_gold_rows` returns. The value is still computed at `snc.py:555-565` and is now
unused; the sole consumer, `_row_needs_settlement` at `snc.py:74`, still reads it. This looks like a
casualty of the `custom_pure_qty` edit in the same hunk rather than an intentional change.

**Why** — `row.get("batch_voucher_type")` now always returns `None`, so the `== "Customer Repair"`
branch can never be taken and the function falls straight through to the ownership tests. The
function's own docstring states the rule this deletes: a batch received under a Customer Repair
voucher is never settled by SNC, because repair gold is returned as-is and no owner gold is owed
against it.

**How** — `_row_needs_settlement` is the single gate for both `create_snc` (`snc.py:181-185`) and
`_mwo_needs_settlement` (which calls it at `snc.py:50`). A Customer Repair row now enters
`settle_rows`, so `stamp_snc_requirement` writes `snc_requirement = "Need"` and blocks the for_fg MWO
submit until SNC is run, and running it emits a Repack plus Metal Conversion pair and a submitted
`Material Transfer (WORK ORDER)` (`snc.py:507-508`) against gold that is meant to be handed back
untouched. That is real ledger movement plus an owner-gold draw that is never owed.

**When** — Any PMO whose operation received a batch with
`custom_customer_voucher_type = 'Customer Repair'`. **This is the default path** — it is not scoped
to the new `include_all_items` mode and needs no new button.

**Where** — `jewellery_erpnext/customer_subcontracting/sub_utils/snc.py:579`

```python
# snc.py:555-565 — the value is still read from the Batch
	inventory_type = batch_customer = batch_voucher_type = None
	if batch_no:
		inventory_type, batch_customer, batch_voucher_type = frappe.db.get_value(
			"Batch", batch_no,
			["custom_inventory_type","custom_customer","custom_customer_voucher_type"],
		) or (None, None, None)

# snc.py:566-582 — but the key is no longer in the dict that is returned
	receive_items.append(
		{
			...
			"customer": batch_customer,
			"batch_customer": batch_customer,
			"custom_pure_qty": flt(qty * _get_item_purity(item_code) / 100, 3) if item_code.startswith("M-G-") else qty,
			"s_warehouse": row.get("s_warehouse"),
		}
	)

# snc.py:63-79 — the sole consumer, untouched by this PR
def _row_needs_settlement(mwo, row, pmo_is_customer_gold):
	"""...
	A batch received under a "Customer Repair" voucher is never settled by SNC,
	regardless of order type or batch ownership (business rule): repair gold is
	returned as-is, so no owner gold is owed against it.
	..."""
	if row.get("batch_voucher_type") == "Customer Repair":
		return False
```

The removal itself, from the diff:

```diff
-				"batch_voucher_type": batch_voucher_type,
```

```mermaid
flowchart TD
  A["Batch with custom_customer_voucher_type 'Customer Repair'"] --> B["_get_receivable_gold_rows reads it\nsnc.py:555-565"]
  B --> C["row dict appended without the key\nsnc.py:566-582"]
  C --> D["_row_needs_settlement sees None\nsnc.py:74"]
  D --> E["falls through to the ownership tests"]
  E --> F["row enters settle_rows\nsnc.py:181-185"]
  F --> G["Repack plus Metal Conversion and a submitted\nMaterial Transfer against repair gold"]
  F --> H["_mwo_needs_settlement true\nsnc_requirement Need, FG MWO submit blocked"]
```

Note that CI cannot catch this: `tests/test_snc_submit_guard.py` patches `_get_receivable_gold_rows`
wholesale (`:144`, `:160`, `:313`, `:520`, `:669`, `:710`) and hand-builds the row dict at `:221-243`
with `"batch_voucher_type": voucher_type` at `:237`, so the producer is never exercised and the two
Customer Repair tests keep passing while production behaviour has flipped.

**Fix**

```diff
 			"customer": batch_customer,
 			"batch_customer": batch_customer,
+			"batch_voucher_type": batch_voucher_type,
 			"custom_pure_qty": ...,
```

Then add a test that calls `_get_receivable_gold_rows` itself (rather than hand-building the dict)
and asserts the key survives, so the next deletion is caught. If the repair rule really was meant to
change, `_row_needs_settlement` and its docstring have to change in the same commit — but nothing
else in this PR suggests that.

---

### 2. New Purchase Invoice `before_validate` queries a Sales Order Item column that does not exist  ·  **BLOCKER**

**What** — `hooks.py:245` newly registers `purchase_invoice.before_validate`, which calls
`update_rate_from_sales_invoice`. That function runs raw SQL against
`tabSales Order Item.custom_design_code`. The field is introduced by this PR and exists in no
custom_fields JSON, no fixture and no patch; `patches.txt` and `install.py` are untouched.

**Why** — `custom_fields/*.json` is inert on its own (`hooks.py:12`, `after_migrate` is commented
out), and a schema-creating patch needs both `patches.txt` and `install.py:_run_schema_patches()`
(`install.py:663-677`). Neither is present here. Raw `frappe.db.sql` does not degrade on an unknown
column — MariaDB raises 1054, out of `before_validate`, so the document cannot be saved at all.
I checked the three live sites directly: `tabSales Order Item.custom_design_code` does not exist on
`alfarsi`, `gk` or `kg-gk`, and it is absent from `custom_fields/sales_order_item.json` (96
fieldnames) and from gke_customization's `fixtures/custom_field.json` (76 Sales Order Item
fieldnames).

**How** — Create a Purchase Invoice from any Purchase Order (so `row.purchase_order` is populated)
and save. `before_validate` → `update_rate_from_sales_invoice` → the SELECT →
`OperationalError (1054) Unknown column 'custom_design_code' in 'where clause'`. Save fails.

**When** — Immediately on deploy, for **every Purchase Invoice with at least one PO-linked item
row** (a PI with no PO-linked row still saves). That is ordinary AP work, not just the job-work
flow, and it is guaranteed on CI's fresh `test_site`.

**Where** — `jewellery_erpnext/jewellery_erpnext/doc_events/purchase_invoice.py:26`

```python
	for row in doc.items:
		if row.purchase_order:
			# Get any Sales Order Item where po_no = purchase_order and custom_design_code is set
			so_item = frappe.db.sql("""
				SELECT parent
				FROM `tabSales Order Item`
				WHERE po_no = %s AND custom_design_code IS NOT NULL AND custom_design_code != ''
				LIMIT 1
			""", (row.purchase_order,), as_dict=True)
```

```python
# hooks.py:243-246 — the registration is new; base registers only `validate`
	"Purchase Invoice": {"validate": [_SUPPLIER_ALLOWED_ITEM_VALIDATOR,
			"jewellery_erpnext.jewellery_erpnext.doc_events.purchase_invoice.update_effective_tax_rate"],
		"before_validate": "jewellery_erpnext.jewellery_erpnext.doc_events.purchase_invoice.before_validate",
	},
```

One side effect worth calling out separately: because the base branch hooks no `before_validate` for
this doctype, `update_expense_account` also starts running on every Purchase Invoice for the first
time. Is that intended, or should the hook be scoped to the job-work case?

**Fix** — Ship the field: add `Sales Order Item.custom_design_code` to
`custom_fields/sales_order_item.json`, write a patch that calls `create_custom_fields` for it, and
register that patch in **both** `patches.txt` and `install.py:_run_schema_patches()` — `bench
install-app` marks patches complete without running them, so a fresh site otherwise never gets the
column. One patch can create this field and the one in issue 3. Until then, gate the whole hook on
the job-work case rather than letting it touch every invoice.

---

### 3. Manufacturing Plan submit joins `soi.custom_parent_manufacturing_order`, also never created  ·  **BLOCKER**

**What** — `process_so_for_employee_ir`'s idempotency query joins `tabParent Manufacturing Order` to
`tabSales Order Item.custom_parent_manufacturing_order`, and the same field is written back at
`:258`. That field ships only for `Stock Entry Detail` and `Stock Entry MOP Item` — in jewellery's
`custom_fields/` and in gke's fixture alike — and I confirmed it is absent from
`tabSales Order Item` on `alfarsi`, `gk` and `kg-gk`.

**Why** — Same inert-JSON / two-place-registration problem as issue 2, but the crash lands at the
*end* of `on_submit` (`manufacturing_plan.py:127`), after the bulk Sales Order Item UPDATE at
`:61-82` and after `create_manufacturing_order` has built the PMO/MWO tree at `:90-99`. The
exception propagates, so the whole submit rolls back and the plan stays in draft — the user watches
the work happen and then loses it.

**How** — Submit a Manufacturing Plan whose rows point at Sales Order Items whose `po_no` resolves to
a Purchase Order with `employee_ir` set. `has_employee_ir` is true at `:169-171`, execution reaches
the SQL, MariaDB raises `Unknown column 'soi.custom_parent_manufacturing_order' in 'on clause'`.

**When** — The first time anyone submits a Manufacturing Plan on this path, on any site where the
field has not been hand-created. Non-job-work plans return before the query, so existing plan
submits will not surface it.

**Where** — `jewellery_erpnext/jewellery_erpnext/doctype/manufacturing_plan/manufacturing_plan.py:177`

```python
		# Idempotency check: verify if a Sales Order is already created for this plan and items
		existing_so = frappe.db.sql("""
			SELECT soi.parent
			FROM `tabSales Order Item` soi
			JOIN `tabParent Manufacturing Order` pmo ON pmo.name = soi.custom_parent_manufacturing_order
			JOIN `tabSales Order` so ON so.name = soi.parent
			WHERE pmo.manufacturing_plan = %s
			AND pmo.sales_order_item IN %s
			AND so.docstatus < 2
			LIMIT 1
		""", (self.name, tuple(so_item_names)))
		if existing_so:
			return
```

```mermaid
flowchart TD
  S["Manufacturing Plan submit"] --> U["bulk UPDATE Sales Order Item\nmanufacturing_order_qty, lines 61 to 82"]
  U --> P["create_manufacturing_order loop\nPMO and MWO tree created, lines 90 to 99"]
  P --> E["create_employee_ir_sales_orders, line 127"]
  E --> Q["idempotency SQL joins soi.custom_parent_manufacturing_order, line 177"]
  Q --> X["MariaDB 1054 unknown column"]
  X --> R["exception leaves on_submit\nthe whole transaction rolls back"]
  R --> Z["plan stays draft and every PMO or MWO above is discarded"]
```

Two related notes on the same hunk:

- `custom_design_code` is written onto the new Sales Order row at `:249`. That column does not exist
  either, so Frappe's `get_valid_dict` drops it silently — and since every row's `item_code` is the
  literal `"Subcontracting Income"`, `custom_design_code` is the only field carrying which design was
  actually made. That data is lost without an error.
- `custom_tracking_bom` at `:254-256` is fine: it does exist on `tabSales Order Item` on all three
  sites as an Administrator-created Custom Field, which is why you will not find it in either app's
  JSON.

**Fix** — Add `Sales Order Item.custom_parent_manufacturing_order` (Link → Parent Manufacturing
Order) and `custom_design_code` to `custom_fields/sales_order_item.json`, plus one patch that
creates both, registered in `patches.txt` **and** `install.py:_run_schema_patches()`. Provisioning
should create and never modify — pass through `create_custom_fields` with that in mind, since its
default `update=True` merges over site customisation. Consider replacing the raw SQL with
`frappe.get_all`/`frappe.db.exists` so a missing field surfaces as a clean Frappe error rather than a
1054.

---

### 4. `make_quotation` now throws for every PO without `custom_pmo` — this is the failing check  ·  **HIGH**

**What** — Two unconditional hard failures were inserted into the PO → Quotation mapper:
`po_doc.supplier` is dereferenced at `:362` for the new Company lookup, and every item row must now
carry `custom_pmo` at `:373-374` or the whole mapping aborts.

**Why** — `.github/workflows/core-tests.yml:65` runs
`jewellery_erpnext.jewellery_erpnext.tests.test_purchase_order`. Its `DummyPO`
(`test_purchase_order.py:10-41`) never assigns `supplier` and defines no `__getattr__`, so
`po_doc.supplier` raises `AttributeError` — patching `frappe.db.get_value` does not help, because the
attribute is evaluated before the call. That hits **all five** `make_quotation` tests
(`:131`, `:210`, `:225`, `:311`, `:413`), including the two that pass `items=[]`; the `custom_pmo`
throw at `:374` is never even reached. Beyond CI, `custom_pmo` on a Purchase Order Item is written by
exactly one place, `EmployeeIR.create_subcontracting_order` (`employee_ir.py:821`), so every PO
raised any other way — including every PO that already exists — can no longer be mapped to a
Quotation at all.

**How** — CI: the Core Tests job errors on the five tests above. Production: a user clicks
Create → Quotation on a PO not raised from an Employee IR and gets
`Row #None: Parent Manufacturing Order (PMO) is required for Quotation mapping.`

**When** — Every CI run on this branch, and every PO → Quotation mapping for a PO without a PMO.

**Where** — `jewellery_erpnext/jewellery_erpnext/doc_events/purchase_order.py:373`

```python
	mapped_company = frappe.db.get_value(
		"Company", {"supplier_code": po_doc.supplier}, "name"
	)
	if mapped_company:
		target_doc.company = mapped_company
	elif not target_doc.company:
		frappe.throw(
			f"Supplier '{po_doc.supplier}' is not mapped to any internal Company. "
			"Please set the 'Supplier Code' field in the appropriate Company record."
		)

	for row in po_doc.items:
		if not row.custom_pmo:
			frappe.throw(f"Row #{row.idx}: Parent Manufacturing Order (PMO) is required for Quotation mapping.")
```

```python
# tests/test_purchase_order.py:10-19 — DummyPO never sets `supplier`
class DummyPO:
	def __init__(self, **kwargs):
		self.name = kwargs.get("name", "PO-1")
		self.purchase_type = kwargs.get("purchase_type", "Finished Goods")
		self.company = kwargs.get("company", "Gurukrupa Export Private Limited")
```

**Fix** — Make both new rules conditional rather than unconditional: read the supplier defensively
(`po_doc.get("supplier")`), and when `row.custom_pmo` is empty fall back to the previous `row.get(...)`
values instead of throwing. If the PMO genuinely must be mandatory, scope it (for example only when
`po_doc.employee_ir` is set) and update the five `make_quotation` tests in the same commit — CI
actually runs that module, so it is the one place where a test edit is not optional.

---

### 5. Customer Goods Issue submit force-completes the MWO, with no reversal on cancel  ·  **HIGH**

**What** — `onsubmit` collects every distinct `custom_manufacturing_work_order` on the item rows of a
"Customer Goods Issue" Stock Entry and writes `status = "Completed"` straight onto each MWO with
`frappe.db.set_value`, in a loop, before `validate_items(self)` runs. No prior state is captured.

**Why** — `Manufacturing Work Order.status` is `read_only: 1`
(`manufacturing_work_order.json:468`), so nothing re-derives it and the operator cannot correct it
from the form. `hooks.py:206` already registers `stock_entry.on_cancel`, and that function
(`stock_entry.py:879-881`) only calls `update_manufacturing_operation` and
`sync_mop_log_for_stock_entry` — this PR does not extend it. The flip is also unconditional per row:
it does not check whether the MWO's operations are actually finished, unlike the two existing writers
of the same transition (`manufacturing_operation.py:573`, `serial_number_creator.py:1090`), which go
through `set_values_in_bulk` from the place that knows.

**How** — Submit a Customer Goods Issue built by the new button (it stamps
`custom_manufacturing_work_order` on every row at `manufacturing_work_order.py:566`), notice a wrong
quantity, cancel it. The MWO is still `Completed`, its previous status is gone, and any report, list
filter or downstream guard keyed on status now treats an in-progress order as done.

**When** — Every Customer Goods Issue submit, and every cancel or amend of one.

**Where** — `jewellery_erpnext/jewellery_erpnext/doc_events/stock_entry.py:935`

```python
def onsubmit(self, method):
	if self.stock_entry_type == "Customer Goods Issue":
		mwos = set()
		for row in self.items:
			if row.get("custom_manufacturing_work_order"):
				mwos.add(row.custom_manufacturing_work_order)
		for mwo in mwos:
			frappe.db.set_value("Manufacturing Work Order", mwo, "status", "Completed")

	validate_items(self)
```

```mermaid
flowchart TD
  A["Customer Goods Issue submit"] --> B["collect custom_manufacturing_work_order from the rows"]
  B --> C["frappe.db.set_value MWO status Completed"]
  C --> D["previous status is never captured"]
  E["Stock Entry cancel"] --> F["on_cancel, stock_entry.py lines 879 to 881"]
  F --> G["update_manufacturing_operation and sync_mop_log_for_stock_entry only"]
  G --> H["MWO stays Completed and status is read_only\nso there is no UI path back"]
```

**Fix** — Add the mirror in `on_cancel`: capture the prior status (or recompute it from the
operations) and restore it for each `custom_manufacturing_work_order` on the cancelled Stock Entry.
Use the single `set_values_in_bulk("Manufacturing Work Order", list(mwos), {"status": "Completed"})`
call the rest of the app uses instead of the per-MWO loop. Best of all, gate the flip on the MWO
actually being complete — a Customer Goods Issue is not by itself proof that the work order is
finished.

---

### 6. Purchase Invoice rates rewritten from an arbitrarily-picked Sales Invoice  ·  **HIGH**

**What** — `update_rate_from_sales_invoice` overwrites `row.rate` and `row.amount` for any row with a
`purchase_order`, on every save. Both lookups are `LIMIT 1`: the first has no `ORDER BY` at all, and
the second accepts draft Sales Invoices (`si.docstatus < 2`) and matches on an `OR` that falls back
to `custom_design_code = row.item_code`.

**Why** — `SELECT parent FROM tabSales Order Item WHERE po_no = %s ... LIMIT 1` with no `ORDER BY`
returns whichever row the storage engine hands back first; when a PO feeds several Sales Order Items
the chosen `so_name` — and therefore the rate — is not reproducible across saves. The second query
then takes the newest matching Sales Invoice *including drafts*, so an unposted draft can set the
purchase rate on a document that is about to be submitted. There is no currency or
`conversion_factor` handling, and no gate on `purchase_type`, company, or the job-work flow.

**How** — A PO covers two Sales Order Items. The buyer saves the PI and gets rate X. Someone edits an
unrelated draft Sales Invoice; the buyer saves again and gets rate Y. Nothing is flagged — the rate
simply changes underneath them, and the invoice is submitted at whatever the last save produced.

**When** — Latent today, because the first query raises 1054 before any rate is written (issue 2).
It becomes live the moment `custom_design_code` exists, and then it applies to every PI row carrying
a `purchase_order`, ordinary purchasing included.

**Where** — `jewellery_erpnext/jewellery_erpnext/doc_events/purchase_invoice.py:42`

```python
			if so_item:
				so_name = so_item[0].parent
				si_rate = frappe.db.sql("""
					SELECT sii.rate 
					FROM `tabSales Invoice Item` sii
					JOIN `tabSales Invoice` si ON si.name = sii.parent
					JOIN `tabSales Order Item` soi ON soi.name = sii.so_detail
					WHERE sii.sales_order = %s 
					  AND si.docstatus < 2
					  AND (soi.custom_po_details = %s OR soi.custom_design_code = %s)
					ORDER BY si.creation DESC
					LIMIT 1
				""", (so_name, row.po_detail, row.item_code))
				# frappe.throw(str(si_rate))
				if si_rate and si_rate[0][0] is not None:
					row.rate = frappe.utils.flt(si_rate[0][0])
					row.amount = frappe.utils.flt(row.qty) * row.rate
```

**Fix** — Gate the function on the job-work case (a `purchase_type` check, or require
`frappe.db.get_value("Purchase Order", row.purchase_order, "employee_ir")`). Restrict the Sales
Invoice query to `si.docstatus = 1`, add a deterministic `ORDER BY si.posting_date DESC, si.name DESC`,
and resolve the Sales Order through `row.po_detail` → the specific Sales Order Item rather than a bare
`LIMIT 1`. Drop the `row.amount` assignment and let `calculate_taxes_and_totals` recompute it at the
right precision, and decide explicitly what happens when the SI and PI currencies differ.

---

### 7. The idempotency guard keys on a nullable link — duplicate submitted Sales Order on amend  ·  **MEDIUM**

**What** — The guard at `:174-183` looks for an existing Sales Order Item whose
`custom_parent_manufacturing_order` joins to a PMO of this plan. But the value written into that
field, `pmo_name` at `:219-223`, is a `frappe.db.get_value` that returns `None` whenever no PMO
exists for `(manufacturing_plan=self.name, sales_order_item=so_item.name)` — and the row is appended
anyway at `:258`.

**Why** — A SQL `JOIN ... ON pmo.name = soi.custom_parent_manufacturing_order` never matches NULL.
The two row sets genuinely differ: `create_manufacturing_order` runs only for rows with
`row.docname or row.mwo` (`:90-99`), while `create_employee_ir_sales_orders` collects the wider
`row.sales_order and row.docname` (`:136-139`). So a plan can legitimately produce Sales Order rows
whose link is NULL, and the guard can never see them.

**How** — Submit such a plan; a Sales Order is created and submitted with NULL in that column. Amend
and re-submit: the guard returns empty and a **second** Sales Order is inserted and submitted at
`:265-268` — duplicate committed subcontracting income and duplicate downstream demand. On an amend
the guard misses for a second, independent reason too: the amended plan has a new `name`, so
`pmo.manufacturing_plan = %s` matches nothing even when every link was written correctly. (A retry
after a rolled-back submit is *not* a duplicate path — the rollback discards the Sales Order as
well.) The back-link that would have repaired the first case is commented out at `:270-280`.

**When** — Any amend or re-submit of an Employee-IR Manufacturing Plan, once issue 3 is fixed and the
query can run at all.

**Where** — `jewellery_erpnext/jewellery_erpnext/doctype/manufacturing_plan/manufacturing_plan.py:219`

```python
			pmo_name = frappe.db.get_value(
				"Parent Manufacturing Order",
				{"manufacturing_plan": self.name, "sales_order_item": so_item.name},
				"name",
			)
```

```python
# :258 — written unguarded
					"custom_parent_manufacturing_order": pmo_name,
```

```mermaid
flowchart TD
  A["row has no PMO for this plan and SO item"] --> B["pmo_name is None, lines 219 to 223"]
  B --> C["Sales Order Item written with a NULL link, line 258"]
  C --> D["Sales Order inserted and submitted, lines 265 to 268"]
  D --> E["amend or re-submit the plan"]
  E --> F["guard JOIN cannot match a NULL, lines 174 to 183"]
  E --> G["amended plan has a new name so pmo.manufacturing_plan misses too"]
  F --> H["a second Sales Order is created and submitted"]
  G --> H
```

**Fix** — Do not key idempotency on a nullable link. Either skip the row when `pmo_name` is falsy
(and throw if that leaves nothing to create), or better, key off the Manufacturing Plan itself: add a
`custom_manufacturing_plan` Link on the Sales Order, shipped by the same patch as the other fields,
and guard with `frappe.db.exists("Sales Order", {"custom_manufacturing_plan": ..., "docstatus": ["<", 2]})`
— and make it survive an amend (match on `amended_from` lineage, or on the source Sales Order Items).
Also drop `ignore_permissions=True` at `:267` or say in a comment why a user who may submit the plan
may not create a Sales Order.

---

### 8. The auto-created Sales Order can be submitted at rate 0  ·  **MEDIUM**

**What** — `rate` defaults to `0.0` and stays 0 whenever the Party Link lookup (`:191-201`) or the
`item_subcategory` lookup (`:226-228`) misses. The Supplier Services Price query has no `ORDER BY`,
no `LIMIT` and no `purchase_type` filter, and the code takes `ssp_rate[0][0]`. The document is then
inserted and submitted from inside Manufacturing Plan `on_submit`.

**Why** — `Supplier Services Price` carries a `purchase_type` field, so one supplier can legitimately
have several price records; without an `ORDER BY` the chosen row is engine order. When `supplier` or
`item_subcategory` is falsy the rate is never looked up at all, and the only user signal is an
`alert=True` msgprint at `:282-287`, which does not block and is easy to miss. The nested `submit()`
also means any Sales Order validation failure rolls the entire Manufacturing Plan submit back.

**How** — A customer with no Party Link, or an item with no `item_subcategory`, produces a submitted
revenue document worth zero that has to be cancelled and amended by hand. Two Supplier Services Price
rows for the same supplier and subcategory but different `purchase_type` make two otherwise identical
plans produce different rates.

**When** — Every Manufacturing Plan submit that reaches the Employee-IR branch, once issue 3 is
fixed.

**Where** — `jewellery_erpnext/jewellery_erpnext/doctype/manufacturing_plan/manufacturing_plan.py:230`

```python
			rate = 0.0
			if supplier and item_subcategory:
				ssp_rate = frappe.db.sql(
					"""
					SELECT child.supplier_fg_purchase_rate
					FROM `tabSupplier Services Price` parent
					JOIN `tabSupplier Services Price Item Subcategory` child
						ON child.parent = parent.name
					WHERE parent.supplier = %s AND child.sub_category = %s
				""",
					(supplier, item_subcategory),
				)
				if ssp_rate and ssp_rate[0][0]:
					rate = frappe.utils.flt(ssp_rate[0][0])
		...
		new_so.set_missing_values()
		new_so.calculate_taxes_and_totals()
		new_so.insert(ignore_permissions=True)
		new_so.submit()
```

**Fix** — Throw (or skip the row with a blocking message) when no rate can be resolved, rather than
submitting at 0. Add `AND parent.purchase_type = %s` and a deterministic
`ORDER BY parent.modified DESC LIMIT 1`, or replace the SQL with `frappe.get_all` on the child table.
Leaving the Sales Order in **draft** for review, instead of submitting it from inside another
document's `on_submit`, would also remove the rollback coupling.

---

### 9. Hardcoded `"Subcontracting Income"` Item that nothing provisions  ·  **MEDIUM**

**What** — Every row of the auto-created Sales Order is written with the literal
`item_code = "Subcontracting Income"`. No fixture, patch or `create_test_data` entry creates that
Item; `git grep` at this head finds the string only here and in a comment in `sales_invoice.py:1476`.

**Why** — Link validation raises `LinkValidationError: Could not find Item: Subcontracting Income` at
`new_so.insert()` (`:267`), inside Manufacturing Plan `on_submit`, so the plan submit fails outright.
This is the same class of unmigrated dependency as the missing custom fields, and the site data
confirms it: the Item exists on **`gk` only** — it is absent on `alfarsi` and on `kg-gk`, and will be
absent on CI's `test_site`.

**When** — Any site where an administrator has not created the Item by hand: fresh installs, CI, any
new company rollout.

**Where** — `jewellery_erpnext/jewellery_erpnext/doctype/manufacturing_plan/manufacturing_plan.py:248`

```python
			new_so.append(
				"items",
				{
					"item_code": "Subcontracting Income",
					"custom_design_code": so_item.item_code,
					"qty": so_item.qty,
```

**Fix** — Make it configurable and validated: read the item from a settings Single rather than
hardcoding it, and `frappe.throw` with a clear message when it is unset or
`frappe.db.exists("Item", ...)` is false, so the failure is an actionable message up front rather
than a LinkValidationError mid-insert. If it must stay hardcoded, create the Item in the same patch
that creates the new custom fields.

---

### 10. The MOP Log query behind Customer Goods Issue filters on the wrong things  ·  **MEDIUM**

**What** — `make_customer_goods_issue` selects MOP Log rows with `"docstatus": ["!=", 2]`, reads
`qty_after_transaction`, and takes the newest row per `item_code` across the whole MWO.

**Why** — Three separate problems, all verifiable at this head. (1) `mop_log.json` has no
`is_submittable` key, so every MOP Log row is `docstatus = 0` and the `!= 2` filter excludes nothing;
the real cancellation marker is `is_cancelled` (`mop_log.json:58`), which this query never consults —
the house convention elsewhere is exactly `{"manufacturing_work_order": mwo, "is_cancelled": 0}`
(`stock_entry.py:1972`). (2) The doctype carries four balance tiers (`mop_log.json:131`, `:152`,
`:163`, `:173`) and `qty_after_transaction_batch_based` is the authoritative one; reading
`qty_after_transaction` gives a diluted figure. (3) `latest_per_item` is keyed on `item_code` alone
(`:547-550`) while the filter spans the entire MWO and `manufacturing_operation` is neither selected
nor filtered, so when an item appears in more than one operation the row that happened to be written
last supplies both the qty and the `s_warehouse` (`:559-560`, `:565`).

**How** — An operator cancels a MOP movement (which by design sets `is_cancelled = 1` and leaves
docstatus alone), then clicks Customer Goods Issue. The cancelled row is the newest for that item, so
its balance is proposed as the issue quantity out of its warehouse. The method returns an *unsaved*
Stock Entry that the browser opens as a draft, so this is a wrong pre-filled proposal rather than an
automatic posting — but it is a plausible-looking one, and submitting it as presented also flips the
MWO to Completed (issue 5).

**When** — Every use of the new button on an MWO with a cancelled MOP Log row or with the same item
in more than one operation.

**Where** — `jewellery_erpnext/jewellery_erpnext/doctype/manufacturing_work_order/manufacturing_work_order.py:540`

```python
		mop_logs = frappe.get_all(
			"MOP Log",
			filters={
				"manufacturing_work_order": self.name,
				"docstatus": ["!=", 2],
				"item_code": ["!=", self.item_code]
			},
			fields=["item_code", "qty_after_transaction", "pcs_after_transaction", "to_warehouse", "creation"],
			order_by="creation desc"
		)
```

```mermaid
flowchart TD
  A["all MOP Log rows for this MWO"] --> B["filter docstatus not equal to 2"]
  B --> C["MOP Log is not submittable, every row is docstatus 0\nso nothing is excluded"]
  C --> D["order by creation desc, first row per item_code wins"]
  D --> E["a row with is_cancelled 1 can win"]
  D --> F["a row from a different manufacturing_operation can win"]
  E --> G["qty, transfer_qty and s_warehouse on the proposed Stock Entry"]
  F --> G
```

**Fix** — Replace the filter with
`{"manufacturing_work_order": self.name, "is_cancelled": 0, "item_code": ["!=", self.item_code]}` and
drop the meaningless `docstatus` clause. Select `qty_after_transaction_batch_based`. Key
`latest_per_item` on `(manufacturing_operation, item_code)` — adding `manufacturing_operation` to
`fields` — or document why an MWO-wide latest row is correct here. `order_by="creation desc, name desc"`
also makes the pick deterministic when two rows share a timestamp.

---

### 11. The Receive gate runs in `validate` and picks the Issue IR non-deterministically  ·  **MEDIUM**

**What** — `validate_purchase_invoice` is called from `validate()` (`employee_ir.py:175`), so its two
throws (`:209-213`, `:215-219`) fire on an ordinary draft save. It resolves the originating Issue IR
with a `LIMIT 1` and no `ORDER BY`, and inspects only `employee_ir_operations[0]`.

**Why** — An operator cannot save a partially-filled Receive IR until finance has posted the invoice,
which is a workflow block rather than a data rule. And `set_repeat_receive_flag`'s own docstring in
this file says an MWO can pass through the same department more than once — so several
`Employee IR Operation` rows can satisfy
`manufacturing_work_order = %s AND parent.type = 'Issue' AND department = %s`, and an unordered
`LIMIT 1` picks an arbitrary one. Only the first child row's MWO is examined, so a multi-MWO receive
is validated against one arbitrary lineage.

**How** — On a repeat receive the query may return the older Issue IR, whose PO already has a
submitted Purchase Invoice, so the gate passes though the current cycle has no invoice — or it
returns the newer one and blocks a receive that is legitimately covered. Either way the outcome is
not reproducible.

**When** — Every save of a subcontracting Receive Employee IR; the non-determinism appears whenever
an MWO has been issued to the same department more than once.

**Where** — `jewellery_erpnext/jewellery_erpnext/doctype/employee_ir/employee_ir.py:177`

```python
			issue_ir = frappe.db.sql("""
				SELECT parent 
				FROM `tabEmployee IR Operation` 
				WHERE manufacturing_work_order = %s 
				AND parent IN (SELECT name FROM `tabEmployee IR` WHERE type = 'Issue' AND department = %s AND docstatus = 1)
				LIMIT 1
			""", (mwo, self.department))
			if issue_ir:
				issue_name = issue_ir[0][0]
				po = frappe.db.get_value("Purchase Order", {"employee_ir": issue_name, "docstatus": ["<", 2]}, "name")
```

**Fix** — Move the gate to a `before_submit` path so drafts stay saveable. Add
`ORDER BY creation DESC` (or join on the specific receive lineage) so the Issue IR choice is
deterministic, and iterate every row of `employee_ir_operations` rather than `[0]`. One more thing to
decide: the `tabPurchase Invoice Item` check matches any submitted row, including one on a return or
credit note — add `is_return = 0` if that is not intended.

---

### 12. Can a Receive still be completed when the Issue IR produced no Purchase Order?  ·  **MEDIUM**

**What** — The gate throws
`"No Purchase Order found for Issue IR {0}. A Purchase Invoice is required before receiving."`
(`employee_ir.py:214-219`) whenever it finds a submitted Issue IR but no Purchase Order linked back
to it. Meanwhile `create_subcontracting_order` — the only producer of that PO — returns without
creating anything when the PO ends up with no item rows (`employee_ir.py:824-825`), and it is invoked
only from the Issue submit path (`employee_ir.py:130-134`).

**Why the question** — On that reading the missing document can never be produced after the fact: the
PO is only ever built during the Issue IR's own `on_submit`, and that Issue IR is already submitted.
Combined with issue 11 (the check runs in `validate`), the Receive IR could not even be saved as a
draft, leaving the work in progress with no UI path forward.

**Can you confirm that an Issue IR can never legitimately reach submit with an empty PO?** If it can,
this needs an escape hatch — treating "no Purchase Order for the Issue IR" as a skip rather than a
hard block would be the simplest one, since a subcontracting Issue that created no PO has no invoice
to require. If it genuinely cannot happen, a one-line comment at the throw saying so would save the
next reader the same trip.

**Where** — `jewellery_erpnext/jewellery_erpnext/doctype/employee_ir/employee_ir.py:214`

```python
# employee_ir.py:824-825 — the only PO producer can silently create nothing
		if not po.items:
			return
```

---

### 13. `set_repeat_receive_flag()` is now inside `validate_purchase_invoice` and gets skipped  ·  **MEDIUM**

**What** — The new method was inserted directly above the existing `self.set_repeat_receive_flag()`
call without re-indenting that call, so it is now the last statement of `validate_purchase_invoice`
(two tabs) instead of the last statement of `validate`. `validate_purchase_invoice` contains two bare
`return`s, at `:180` and `:185`, plus the two throws.

**Why** — For a Receive IR with `subcontracting = "Yes"` whose `employee_ir_operations` table is empty,
or whose first row has no `manufacturing_work_order`, the method returns before line 220 and the flag
is never recomputed. Its own docstring (`:222-231`) explains why that matters: the value is
"Recomputed from the database rather than trusted from the client … the flag drives a `depends_on`, so
a posted value could otherwise reveal (or hide) the field at will." Skipping it means a client-posted
`is_repeat_receive` / `worker_performance` survives onto the document. (To be precise: the
`type != "Receive"` clearing branch at `:232-235` is *not* affected — for a non-Receive IR the outer
`if` at `:178` is false and line 220 runs normally.)

**When** — Any subcontracting Receive Employee IR saved with an empty or MWO-less operations table.
It is also a structural hazard: any future `return` added to `validate_purchase_invoice` silently
disables the recompute for more cases.

**Where** — `jewellery_erpnext/jewellery_erpnext/doctype/employee_ir/employee_ir.py:220`

```python
	def validate_purchase_invoice(self):
		if self.type == "Receive" and self.subcontracting == "Yes":
			if not self.employee_ir_operations:
				return

			# Trace back to the Issue IR using the first MWO in the child table
			mwo = self.employee_ir_operations[0].manufacturing_work_order
			if not mwo:
				return
			...
		self.set_repeat_receive_flag()
```

**Fix** — De-indent line 220 by one tab so it is a sibling of `self.validate_purchase_invoice()` in
`validate()`, and delete it from the end of `validate_purchase_invoice`:

```diff
 		self.validate_fg_bom_fields()
 		self.validate_purchase_invoice()
+		self.set_repeat_receive_flag()
```

---

### 14. `include_all_items` puts carats into `custom_pure_qty`, then hard-throws on the same rows  ·  **MEDIUM**

**What** — The new mode admits `F-G`, `D-` and `G-` prefixed items alongside `M-G-` (`:537-550`), and
for all of them writes `custom_pure_qty = qty` verbatim (`:579`). Back in `create_snc`, any admitted
row that does not find an owner batch hits a new unconditional `frappe.throw` at `:223-226`.

**Why** — `custom_pure_qty` is a fine-metal weight everywhere else in this module: it is copied onto
transfer rows at `:215` and onto the Stock Entry Detail at `:496`, and passed as `required_pure_qty`
at `:254` and `:273`. Setting it equal to the raw quantity for a `D-` row means a **carat** count
lands in a field whose consumers read grams of fine metal — and `create_material_transfer_work_order`
inserts and submits that Stock Entry (`:507-508`), so the mixed unit reaches a posted document. This
is the same class of defect as the known fine-versus-reference gram mixing in settlement. Separately,
the new throw aborts the entire settlement mid-way — after the gold rows in `settle_rows` have
already been processed — whenever the owner holds no stock of that non-gold item.

**When** — Any use of "Create SNC (All Materials)" on an MWO with non-gold receivable rows. The
ternary at `:579` is on the shared function, but on the default path the else branch is unreachable
because non-gold rows are filtered out at `:540-542`.

**Where** — `jewellery_erpnext/customer_subcontracting/sub_utils/snc.py:579`

```python
		if not include_all_items:
			if not item_code.startswith("M-G-"):
				continue
		else:
			if not (
				item_code.startswith("M-G-")
				or item_code.startswith("F-G")
				or item_code.startswith("D-")
				or item_code.startswith("G-")
			):
				continue
```

```python
# snc.py:579
				"custom_pure_qty": flt(qty * _get_item_purity(item_code) / 100, 3) if item_code.startswith("M-G-") else qty,

# snc.py:221-226
			continue

		if not row["item_code"].startswith("M-G-"):
			frappe.throw(
				_("No available stock for {0} (Item: {1}).").format(_owner_label(owner_customer), row["item_code"])
			)
```

**Fix** — Do not overload `custom_pure_qty`: leave it 0 or `None` for non-metal rows (and have
`_append_item` omit the field then), or compute a real per-class pure quantity. Either way, state the
unit contract in the docstring of `_get_receivable_gold_rows`, since two callers now depend on it.
Replace the unconditional throw with a `continue` that collects unsettleable rows and reports them
once at the end, or at minimum make the message say the real reason — "not gold, cannot be settled by
SNC" — rather than "No available stock".

---

### 15. 415 lines across 9 files, zero tests, zero workflow registration  ·  **MEDIUM**

**What** — The PR adds no `test_*.py` and edits none of
`.github/workflows/{core,inventory,manufacturing,refining}-tests.yml`. Every new behaviour ships
untested: `create_employee_ir_sales_orders` / `process_so_for_employee_ir`, `make_customer_goods_issue`,
the Customer Goods Issue branches in `stock_entry`, `validate_purchase_invoice`,
`update_rate_from_sales_invoice`, the `include_all_items` SNC mode, and the PMO/company remapping in
`make_quotation`.

**Why** — CI here is a curated allowlist, so an unregistered test never runs — but the inverse bites
too. The modules this PR touches **are** allowlisted, so once issue 4 is fixed those suites go green
and will be read as coverage of the new code when none of the new branches is exercised. Concretely:
issue 1 sails past `test_snc_submit_guard` because that module patches `_get_receivable_gold_rows`
wholesale and hand-builds the row dict, and `validate_purchase_invoice` is never reached because the
Employee IR suite has no `type="Receive"` + `subcontracting="Yes"` case (`subcontracting = "Yes"`
appears only at `test_employee_ir.py:559` and `:597`, both Issue tests).

**Where** — repo-wide; see the Tests & CI section below for the exact allowlist lines.

**Fix** — At minimum: (1) `process_so_for_employee_ir` — the no-Employee-IR-PO early return, the
`pmo_name is None` path, and re-submit idempotency; (2) `validate_purchase_invoice` on a Receive +
subcontracting IR both with and without a submitted PI, plus one asserting `set_repeat_receive_flag`
still ran; (3) a test calling `_get_receivable_gold_rows` directly that asserts `batch_voucher_type`
survives; (4) `make_customer_goods_issue` with an `is_cancelled=1` MOP Log row present. Register every
new module in the matching workflow **in this PR** — an unregistered test file never runs.

---

### 16. `include_all_items` arrives from the client as a string  ·  **LOW**

**What** — `create_snc` is `@frappe.whitelist()`, so the argument reaches Python as a string over
REST, and it is used directly in boolean context without `cint()`. `"0"` and `"false"` are both
truthy.

**Why** — Frappe passes form-encoded arguments through as strings, and the parameter has no type
annotation, so nothing coerces it. The app already knows this: `cint(getattr(mwo, "snc_done", 0))` at
`snc.py:22`, and `is_subcontracting_selected` at `manufacturing_plan.py:23-30` exists precisely
because "the string 'No' is truthy". The new button sends `1`, so the shipped path is fine — the gap
is for every other caller.

**Where** — `jewellery_erpnext/customer_subcontracting/sub_utils/snc.py:164`

```python
@frappe.whitelist()
def create_snc(mwo, include_all_items=False):
```

**Fix** — `include_all_items = cint(include_all_items)` as the first line of `create_snc` (`cint` is
already imported at `snc.py:6`). While you are there, consider moving the `M-G-` / `F-G` / `D-` / `G-`
prefix list into a module constant next to `PURITY_PRIORITY` so the two prefix tests cannot drift.

---

### 17. Three round trips per row plus a nested Sales Order insert and submit inside `on_submit`  ·  **LOW**

**What** — `create_employee_ir_sales_orders` loops over every distinct source Sales Order
(`:144-145`) and, for each, `process_so_for_employee_ir` loops over `parent_so.items` issuing three
separate round trips per row — a PMO `get_value` (`:219-223`), an Item `get_value` (`:226-228`) and
the raw Supplier Services Price query (`:232-241`) — then runs `set_missing_values`,
`calculate_taxes_and_totals`, `insert` and `submit` (`:265-268`). All of it runs synchronously inside
Manufacturing Plan `on_submit`, on top of an `on_submit` that has just built a PMO/MWO tree for every
plan row. There is no `frappe.enqueue` anywhere in the new code.

**Why** — Query count is 3 × rows per source Sales Order, and each nested `insert` + `submit` drags
the full Sales Order hook chain along inside the plan's own write transaction, holding its row locks
for the duration. If anything in that chain raises, the whole plan submit rolls back after the
PMO/MWO work is done.

**Where** — `jewellery_erpnext/jewellery_erpnext/doctype/manufacturing_plan/manufacturing_plan.py:213`

**Fix** — Hoist the per-row lookups: one `frappe.get_all("Parent Manufacturing Order", filters={"manufacturing_plan": self.name, "sales_order_item": ["in", so_item_names]}, fields=["name","sales_order_item"])`
into a dict, one `frappe.get_all("Item", filters={"name": ["in", item_codes]}, fields=["name","item_subcategory"])`,
and one Supplier Services Price query with `child.sub_category IN %(subcats)s`. Then consider moving
the Sales Order creation off the request path —
`frappe.enqueue(..., queue="long", job_id=f"eir-so::{self.name}", deduplicate=True, enqueue_after_commit=True)`,
matching the `job_id` + `deduplicate` pattern every other `frappe.enqueue` in this app uses.

---

### 18. Small cleanups  ·  **NIT**

None of these change behaviour (CI runs no linter), but they are cheap:

```python
# manufacturing_work_order.py:553-575 — import inside the loop body, and an underscore attribute
		for item_code, log in latest_per_item.items():
			from frappe.utils import flt
			if flt(log.qty_after_transaction) > 0:
		...
		se._customer = pmo.customer if pmo else None
		if hasattr(se, "customer"):
			se.customer = pmo.customer if pmo else None

# purchase_invoice.py:45 — leftover debug line
				# frappe.throw(str(si_rate))

# purchase_order.py:374 — untranslated user-facing message
			frappe.throw(f"Row #{row.idx}: Parent Manufacturing Order (PMO) is required for Quotation mapping.")
```

Hoist the `flt` import to module scope, delete the commented-out throw, wrap the new user-facing
throws in `_()`, and drop the stray blank line added inside the `doc_events` dict at `hooks.py:154`.
`se._customer` is the one that reads as if it does something: `get_valid_dict` will never persist a
leading-underscore attribute, and paired with the `hasattr(se, "customer")` guard on the next line it
is unclear which of the two is load-bearing — a one-line comment or a deletion would settle it.

---

## Tests & CI

- **No test file is added or changed by this PR**, and no `.github/workflows/*-tests.yml` is touched.
  The test diff is empty; the whole 415 lines are 9 source files.
- The red check is issue 4, in the allowlisted module
  `jewellery_erpnext.jewellery_erpnext.tests.test_purchase_order`
  (`.github/workflows/core-tests.yml:65`). All five `make_quotation` tests fail.
- The suites that already cover the touched modules and will go green **without exercising any new
  code**: `--doctype "Employee IR"` (`core-tests.yml:52`), `test_purchase_order`
  (`core-tests.yml:65`), `test_purchase_invoice` (`core-tests.yml:66`),
  `--doctype "Manufacturing Plan"` (`manufacturing-tests.yml:50`),
  `--doctype "Manufacturing Work Order"` (`manufacturing-tests.yml:52`), `test_snc_submit_guard`
  (`inventory-tests.yml:57`).
- Two specific blind spots: `test_snc_submit_guard.py` patches `_get_receivable_gold_rows` at `:144`,
  `:160`, `:313`, `:520`, `:669` and `:710` and hand-builds rows at `:221-243`, so issue 1 is
  structurally invisible to it; and
  `jewellery_erpnext/jewellery_erpnext/doctype/employee_ir/test_employee_ir.py` has no
  Receive + `subcontracting = "Yes"` case at all.
- Nothing to flag on rollback hygiene, since no tests were added — but for the ones you add: a
  DB-writing class with `def setUpClass: pass` leaks onto the site permanently (rollback is
  registered inside `setUpClass`), rollback is class-scoped rather than per-test, and a
  `frappe.db.commit()` inside a test escapes it for good.

This PR sits on three of the most contended files in the current set:

```mermaid
flowchart LR
  P["PR 1043"] --> H["jewellery_erpnext/hooks.py"]
  P --> S["doc_events/stock_entry.py"]
  P --> M["doctype/manufacturing_work_order/manufacturing_work_order.py"]
  H --> H2["also 660, 1045, 1115, 1201, 1251, 1260, 1272, 1275, 1276"]
  S --> S2["also 1115, 1201, 1237, 1260, 1275, 1276"]
  M --> M2["also 1201, 1233, 1240, 1244, 1260"]
```

- `hooks.py` is edited by 10 open PRs, `doc_events/stock_entry.py` by 7 and
  `manufacturing_work_order.py` by 6. Your edits to all three are small and additive, so the merge
  risk is textual conflict rather than semantic collision — but the `hooks.py` Purchase Invoice block
  is one others also rewrite, so rebase rather than merge when the time comes.
- The cross-PR merge ordering puts this PR in the last group, with #1045 and #1260, as "resolve or
  close" — purely because of the red check. Fixing issue 4 moves it out of that group.
- If any of the schema-carrying PRs (#660, #1275, #1276) lands first, re-check your new custom field
  names against what they provision before choosing them: #660 alone carries 2,134 Custom Field
  records across 179 DocTypes, 91 of them on Sales Order Item.

## Suggested order of work

1. **Restore `"batch_voucher_type"` in `snc.py`** (issue 1) — one line, and it is an existing
   production behaviour this PR silently turns off. Worth splitting into its own commit.
2. **Fix the red check** (issue 4): make the supplier read and the PMO requirement conditional, and
   update the five `make_quotation` tests in the same commit.
3. **Ship the two missing Sales Order Item fields** via one patch registered in `patches.txt` and
   `install.py:_run_schema_patches()` (issues 2 and 3), and provision the `"Subcontracting Income"`
   Item or make it configurable (issue 9). Until this is done, nothing on the Manufacturing Plan or
   Purchase Invoice path can run anywhere but `gk`.
4. **Close the lifecycle gaps**: MWO status reversal on cancel (issue 5), the NULL-keyed idempotency
   guard (issue 7), and the zero-rate submit (issue 8).
5. **Tighten the queries**: MOP Log `is_cancelled` + batch-based tier + per-operation key (issue 10),
   the Sales Invoice rate lookup (issue 6), the Issue IR pick (issue 11).
6. **Re-indent `set_repeat_receive_flag()`** (issue 13) and settle the `custom_pure_qty` unit contract
   (issue 14) and the `cint` coercion (issue 16).
7. **Add tests and register them** in the matching workflow (issue 15), then the nits (issue 18).

Happy to go deeper on any of these — particularly the schema patch shape or the SNC settlement path,
if that would help.

PR #1045, opened 2026-08-04
Do Not Merge

DHINESH00:


DHINESH00:
## Review — #1045 Do Not Merge

&nbsp;·&nbsp; 0 blocker · 0 high · 6 medium · 7 low
&nbsp;·&nbsp; base `v16_develop_aerele` · head `a4926d45` · 23 files, +4116/-1834

The title already says the intent, so this review is about **what to salvage and how to split it**, not whether to click merge. Short answer: five of the eight commits are independently mergeable today, two need a fix first, and the ruff-format commit should land alone and first so the rest is reviewable.

---

## What this PR does

Eight commits: one project-wide `ruff-format` pass over seven files, and seven independent remediation changes — `eval()` removed from Gemstone Conversion in favour of a strict regex size parser, nine value-interpolating `frappe.db.sql` sites converted to `%(name)s` binding, the dead first element of the two-entry `override_doctype_class["Stock Reconciliation"]` list removed, the QC `force_approve` recursion turned into a bounded iterative walk that no longer re-enters `on_submit`, permission gates on four whitelisted `job_card.py` endpoints, removal of the universal `utils.db_get_value` read primitive plus `write` gates across Refining Entry, and six new test modules.

A lot of this is genuinely good work and should be said plainly:

- **The reformat is provably behaviour-preserving.** An AST comparison of base against `7b85845f` across all seven reformatted files shows one difference in the whole ~4,000-line reflow: docstring indentation at `sales_order.py:2197`. The +1423/-980 and +937/-641 on `sales_order.py` / `sales_invoice.py` are noise, not a rewrite.
- **All six new test modules are registered in `.github/workflows/`** — `core-tests.yml` gets `test_gemstone_size_parser`, `test_sql_parameterisation`, `test_universal_primitives`; `inventory-tests.yml` gets `test_stock_reconciliation_override`; `manufacturing-tests.yml` gets `test_qc_lifecycle` and `test_job_card_endpoint_permissions`. That is the repo's hardest mechanical criterion and most PRs in this batch skip it.
- **The `override_doctype_class` note is exactly right.** Frappe resolves `class_overrides[doctype][-1]` (`frappe/model/base_document.py:117`), `extend_doctype_class` really does exist in this v16, and the first element of the old list really was dead.
- **`db_get_value` really did have exactly one caller** in the whole bench, and the replacement is permission-checked. The docstring's note about `_` being a single-character wildcard in SQL `LIKE` (and the ~137 phantom callers an unescaped search reports) is the kind of detail that saves the next person an hour.
- **The `%%IGST%%` doubling under parameter binding is handled correctly**, and the AST sweeps in the new tests generalise the defect class instead of pinning known instances — run against this tree they report zero offenders, and the gemstone parser round-trips its own 314-value sample with every payload rejected.

```mermaid
flowchart TD
  A["8 commits on remediation base"] --> B["7b85845f ruff-format over 7 files"]
  A --> C["1638f4fe remove eval in Gemstone Conversion"]
  A --> D["178e2ed8 parameterise 9 frappe.db.sql sites"]
  A --> E["6a0020b3 collapse Stock Reconciliation override list"]
  A --> F["a42a001b bound QC force_approve"]
  A --> G["fd7059a6 gate 4 job_card endpoints"]
  A --> H["6d9d63f6 drop utils.db_get_value, gate Refining Entry"]
  A --> I["82517d7c register 6 test modules in CI"]
  B --> J["behaviour preserving reflow"]
  C --> K["independently mergeable as-is"]
  D --> K
  I --> K
  E --> L["needs a scope decision first"]
  F --> L
  G --> M["gate present but names the wrong DocType"]
  H --> M
```

---

## Before → after

The change with the widest behavioural reach is the Stock Reconciliation override consolidation. It is framed in the commit message as "so `difference_amount` is computed", but it also un-stubs upstream row trimming for every ordinary reconciliation on this deployment.

```mermaid
flowchart TD
  subgraph BASE["Base"]
    A1["StockReconciliation.validate"] --> A2["remove_items_with_no_change"]
    A2 --> A3["bare return, validation disabled"]
    A3 --> A4["every typed row is saved, difference_amount left as the form sent it"]
  end
  subgraph HEAD["This PR"]
    B1["StockReconciliation.validate"] --> B2["remove_items_with_no_change"]
    B2 --> B3{"keep_all_rows"}
    B3 -->|"MWO row or custom_auto_creation"| B4["call super, swallow the empty error, restore the rows"]
    B3 -->|"ordinary Desk reconciliation"| B5["upstream super runs unmodified"]
    B5 --> B6["unchanged rows dropped with a msgprint"]
    B5 --> B7["all-unchanged save refused with EmptyStockReconciliationItemsError"]
  end
```

---

## Issues at a glance

| 1 | **MEDIUM** | Consolidation silently enables upstream row trimming and the empty-items throw for every ordinary reconciliation | `stock_reonciliation.py:181` | Every save of a non-MWO, non-auto Stock Reconciliation |
| 2 | **MEDIUM** | `create_internal_transfer` gates on Stock Entry but creates a Job Card Internal Transfer | `job_card.py:321` | Every call; and it locks out Job Card operators without Stock Entry create |
| 3 | **MEDIUM** | `scan_scrap_qr_action` is the one Refining Entry mutator left ungated, and the guard test cannot see it | `refining_entry.py:1594` | Any authenticated session, via `run_doc_method` |
| 4 | **MEDIUM** | Five untimed cross-site `requests.post` calls survive in the `before_validate` path (pre-existing) | `sales_order.py:1227` | Any Sales Order save while the KGGK site is degraded |
| 5 | **MEDIUM** | 5 of 7 `test_qc_lifecycle` tests self-skip on CI's fresh site | `test_qc_lifecycle.py:47` | Every CI run, on every PR |
| 6 | **MEDIUM** | Permission guards are asserted by source-text substring; no test calls a guarded endpoint unprivileged | `test_job_card_endpoint_permissions.py:49` | Any later edit that keeps the guard text but neuters it |
| 7 | **LOW** | `create_new_job_card` gates on Job Card create but mutates a submitted Work Order with permissions off | `job_card.py:151` | Whenever the requested operation is absent from the Work Order |
| 8 | **LOW** | `force_approve` terminates a `duplicate_qc` cycle silently, contrary to the comment above it | `qc.py:132` | Whenever a `duplicate_qc` cycle exists |
| 9 | **LOW** | SQL guard only detects quote-adjacent interpolation, so unquoted value splices pass | `test_sql_parameterisation.py:72` | The next f-string SQL in a numeric or identifier position |
| 10 | **LOW** | Swallowed `EmptyStockReconciliationItemsError` may still show a red error on a save that succeeded | `stock_reonciliation.py:189` | MWO / auto lane saves where no row differs |
| 11 | **LOW** | The new `custom_auto_creation` branch may guard a lane nothing can enter | `stock_reonciliation.py:181` | Always |
| 12 | **LOW** | Three assertions pin the source text of frappe internals | `test_job_card_endpoint_permissions.py:124` | The next frappe bump that touches `_save` / `_submit` |
| 13 | **LOW** | Redundant per-row `frappe.db.get_value` inside the Stock Entry item loop | `material_request.py:71` | Every MR item with a blank warehouse |

---

### 1. Consolidation turns on upstream row trimming and the empty-items throw for every ordinary reconciliation · **MEDIUM**

**What** — Because `keep_all_rows` is False for every Stock Reconciliation that is neither MWO-flagged nor `custom_auto_creation`, line 184 now runs upstream's `remove_items_with_no_change()` unmodified for all of them. Before this PR the live controller's method was a bare `return`, so no Stock Reconciliation on this deployment has ever trimmed rows or raised `EmptyStockReconciliationItemsError`.

**Why** — Upstream's implementation is reached from `StockReconciliation.validate()` at `erpnext/stock/doctype/stock_reconciliation/stock_reconciliation.py:78`, i.e. on every **save**, not just at submit. It ends by throwing when nothing changed and by deleting the unchanged rows (with only a `msgprint`) when some did.

**How** — A stock user enters 20 counted items of which 6 happen to match the system balance exactly, and saves. Under base all 20 rows persist. Under this PR the 6 matching rows are dropped from `self.items`, indexes are renumbered, and the saved document has 14 rows. If all 20 matched, the save is refused outright.

**When** — Every save of a Stock Reconciliation with no MWO-flagged row and no `custom_auto_creation` — the ordinary Desk flow, i.e. the majority. Invisible in CI: `test_stock_reconciliation_override.py` is a `UnitTestCase` of `get_controller` / `__mro__` / `inspect.getsource` assertions and never saves a document.

**Where** — `jewellery_erpnext/jewellery_erpnext/customization/stock_reconciliation/stock_reonciliation.py:181`

```python
		keep_all_rows = self.has_custom_mwo() or bool(self.get("custom_auto_creation"))

		if not keep_all_rows:
			return super().remove_items_with_no_change()
```

what upstream then runs, `erpnext/stock/doctype/stock_reconciliation/stock_reconciliation.py:562-570`:

```python
			frappe.throw(
				_("None of the items have any change in quantity or value."),
				EmptyStockReconciliationItemsError,
			)

		elif len(items) != len(self.items):
			self.items = items
			self.change_idx = True
```

**Fix**
Either state the scope in the commit message and get explicit sign-off that ordinary reconciliations may now drop rows and refuse no-change saves, or keep the old tolerance while still getting the totals — which is what the commit says it wants:

```diff
-		if not keep_all_rows:
-			return super().remove_items_with_no_change()
-
 		original_items = list(self.items)
 		try:
 			super().remove_items_with_no_change()
 		except EmptyStockReconciliationItemsError:
-			pass
+			if not keep_all_rows:
+				raise
 		finally:
-			if len(self.items) != len(original_items):
+			if keep_all_rows and len(self.items) != len(original_items):
 				self.items = original_items
```

Either way, add one `IntegrationTestCase` that actually saves a plain Stock Reconciliation with a mix of changed and unchanged rows and asserts the resulting row count and `difference_amount`. The current module proves only that the source text says `super()`.

Separately: the docstring quantifies existing corruption — "27 of 29 submitted Stock Reconciliations carried 0" `difference_amount` and "922 of 2,450 submitted item rows carried a zero `current_valuation_rate`" — and the PR adds no `patches.txt` entry. That is a scope call rather than a defect, but future reports summing `difference_amount` will mix pre-fix zeros with post-fix real values and nothing records which is which. Worth one sentence in the docstring either way.

---

### 2. `create_internal_transfer` gates on "Stock Entry" but creates a Job Card Internal Transfer · **MEDIUM**

**What** — The gate added to `create_internal_transfer` checks `frappe.has_permission("Stock Entry", "create")`, but the endpoint never touches Stock Entry: it creates and submits a **Job Card Internal Transfer**, raw-writes a Job Card status, and on the last-operation branch deletes Job Card and Work Order Operation rows. The comment above it is copy-pasted from the two sibling endpoints and is factually wrong for this one.

**Why** — `frappe.has_permission(doctype, ptype)` checks only the DocType you name. Naming "Stock Entry" authorises something unrelated to everything this endpoint actually does. Note this cuts **both** ways, and the second direction is the one this PR introduces: at base there was no `has_permission` anywhere in `job_card.py`, so the PR strictly narrows access — but it narrows it along the wrong axis, and a Job Card operator who does not hold Stock Entry create now gets a `PermissionError` on a Job Card workflow they previously used.

**How** — For the mis-gate: a caller with Stock Entry create passes a check that says nothing about the Job Card Internal Transfer they are about to submit or the Job Card row they are about to rewrite. For the lockout: a shop-floor role scoped to Job Cards is refused at line 321 before anything else runs.

To be precise about one thing this does *not* enable: the mass-deletion escalation does not hold. `_delete_remaining_job_cards` enumerates with `frappe.db.get_list`, which is an alias for `frappe.get_list` and is permission-checked, so a caller without Job Card read raises there and the whole request — including the just-submitted transfer — rolls back.

**When** — Every call to this endpoint, in both directions.

**Where** — `jewellery_erpnext/jewellery_erpnext/doc_events/job_card.py:321`

```python
	# These three endpoints create AND submit a Stock Entry. Whitelisting makes a
	# method callable by any authenticated session, including a portal user with no
	# Desk roles, so authorisation has to be explicit. Pattern copied from
	# doc_events/warehouse_stock_entry.py.
...
	frappe.has_permission("Stock Entry", "create", throw=True)
	if balance_gross < gross_wt:
		return frappe.throw(_("Gross Weight Greater Than Balance Gross !"))
	doc = frappe.new_doc("Job Card Internal Transfer")
```

The two siblings at `:220-221` and `:270-271` do match the document they create — this one is the odd one out.

```mermaid
flowchart TD
  A["POST create_internal_transfer"] --> B["has_permission Stock Entry create"]
  B -->|"caller holds Stock Entry create"| C["new_doc Job Card Internal Transfer"]
  C --> D["db.set_value on Job Card status"]
  D --> E{"to_job_card or to_combine_job_card set"}
  E -->|"no, last operation branch"| F["_delete_remaining_job_cards"]
  F --> G["db.get_list is permission checked and throws for a caller without Job Card read"]
  B -->|"Job Card operator without Stock Entry create"| H["PermissionError on a Job Card workflow that used to work"]
```

**Fix**

```diff
-	frappe.has_permission("Stock Entry", "create", throw=True)
+	frappe.has_permission("Job Card Internal Transfer", "create", throw=True)
+	frappe.has_permission("Job Card", "write", throw=True)
 	if balance_gross < gross_wt:
```

and fix the three-line comment above it so it names the document this endpoint really produces. Inside `_delete_remaining_job_cards`, consider `frappe.delete_doc("Job Card", jc.name)` instead of the two `frappe.db.delete` calls so the deletions go through link checks and leave a Deleted Document trail. Then tighten `test_every_stock_creating_endpoint_checks_permission` so it asserts the expected DocType per endpoint rather than the bare substring — as written, that test cements the wrong gate (see issue 6).

---

### 3. `scan_scrap_qr_action` is the one Refining Entry mutator left ungated, and the guard test cannot see it · **MEDIUM**

**What** — Commit `6d9d63f6` states "All eleven now require write". An AST walk of the file at head finds one whitelisted method that mutates `self` and has no gate: `scan_scrap_qr_action`. Its two direct siblings both received `self.check_permission("write")`.

**Why** — `run_doc_method` reaches doc-bound whitelisted methods through `frappe.get_doc(..., check_permission=True)` (`frappe/handler.py:287,291`) and `Document.check_permission()` defaults to `"read"` — the PR's own analysis is correct on this. Refining Entry carries a `{"read": 1, "role": "All"}` permission row, which the PR deliberately keeps and asserts on, so read is the only barrier. The new guard test decides "does this method mutate?" by substring-matching five tokens in the function source; `scan_scrap_qr_action` mutates only via `self.append("material_items", ...)` at `:1617` and `:1649`, none of which contain any of those tokens, so the test classifies it as non-mutating and passes with the hole open.

**How** — A caller with read only invokes it through `run_doc_method` with a chosen `barcode`. The method resolves the batch, runs `_is_blocked_customer_batch` and the FIFO allocation against the warehouse on the supplied document, appends the row and returns the mutated doc. Nothing is persisted — `run_doc_method` never saves — so the concrete harm is an **information oracle** (batch existence, item code, blocked-customer status, per-batch FIFO quantity for an attacker-chosen warehouse) plus an echoed in-memory child row, not a ledger write. That is still the exposure class the commit says it closed, and the inconsistency with three hardened siblings is the more important half.

**When** — Any authenticated session, including website users, on any site where the shipped `role: All` read row is present.

**Where** — `jewellery_erpnext/refining/doctype/refining_entry/refining_entry.py:1594`

```python
	def scan_mwo_action(self, barcode):
		self.check_permission("write")
...
	def scan_serial_no_action(self, barcode):
		self.check_permission("write")
...
	@frappe.whitelist()
	def scan_scrap_qr_action(self, barcode):
		batch = frappe.db.get_value(
			"Batch", {"name": barcode}, ["name", "item"], as_dict=True
		)
```

the detector that cannot see it, `jewellery_erpnext/jewellery_erpnext/tests/test_universal_primitives.py:158-166`:

```python
			mutates = any(
				tok in body
				for tok in (
					".insert(",
					".submit(",
					".save(",
					"db_set(",
					"set_value(",
				)
```

Worth noting the same blind spot almost swallows the two methods that *were* gated: both contain `.save(` only as a commented-out line, so deleting that dead comment would make the guard blind to them too.

```mermaid
flowchart TD
  A["Authenticated session, website user included"] --> B["run_doc_method"]
  B --> C["get_doc with check_permission true, which defaults to read"]
  C --> D["Refining Entry permission row read 1 role All"]
  D --> E{"which scanner is called"}
  E -->|"scan_mwo_action"| F["check_permission write refuses"]
  E -->|"scan_serial_no_action"| F
  E -->|"scan_scrap_qr_action"| G["no gate, method runs"]
  G --> H["batch existence, item code, blocked-customer status and FIFO qty echoed back"]
```

**Fix**

```diff
 	@frappe.whitelist()
 	def scan_scrap_qr_action(self, barcode):
+		self.check_permission("write")
 		batch = frappe.db.get_value(
```

and widen the guard so it cannot miss this class again: add `"self.append("`, `"self.set("` and `"self.extend("` to the token tuple, or better, replace the substring heuristic with an AST walk that flags any call whose func is an attribute on `self` in `{append, set, extend, insert, save, submit, db_set}` plus any `frappe.db.set_value` / `frappe.db.delete`. Re-running the widened guard should report this method red before the fix and green after.

---

### 4. Five untimed cross-site `requests.post` calls survive in the `before_validate` path · **MEDIUM**

**What** — `sales_order.py` is the file this PR reflows most heavily, and five of the six live `requests.post` calls to the KGGK site still pass no `timeout`. Only the one at `:935` carries `timeout=30`. Base has exactly the same 5-of-6 split, so **this PR neither introduces nor worsens it** — it reformats the offending lines and leaves them.

**Why** — `requests.post` with no timeout blocks indefinitely on a hung peer; `requests` has no default. These calls sit in `_get_finding_sub_info` (def `:1213`), `_process_gemstone_detail` (def `:1470`, loop at `:1488`) and `_process_diamond_detail` (def `:1902`, loop at `:1927`), all reached from the `before_validate` hook at `hooks.py:125-129` → `before_validate` (`:42`) → `create_new_bom1(self)` (`:53`) → `_process_single_row` (`:2735`) → the four `_process_*` helpers (`:2623-2626`). So they execute inside the user's save transaction with row locks held, and three of them run once per child row.

**How** — The remote site accepts the connection and stalls. A Sales Order save with 8 diamond rows issues 8 sequential posts that never return, the worker is pinned, and the open transaction holds locks the whole time. With `gunicorn_workers: 17`, a sustained stall exhausts the pool.

**When** — Any Sales Order save (including draft saves) on a company where `Data Migration in KGGK.from_site_1` is set — `use_api = bool(from_site)`.

**Where** — `jewellery_erpnext/jewellery_erpnext/doc_events/sales_order.py:1227`

```python
  1227:		response = requests.post(url=url, headers=headers, data=data)
  1237:			response = requests.post(url, headers=headers, data=payload)
  1532:				response = requests.post(
  1533:					url=f"{from_site}/api/method/gke_customization.gke_order_forms.doc_events.item.get_fixed_retail_rate",
  1615:				response = requests.post(
  1616:					url=f"{from_site}/api/method/gke_customization.gke_order_forms.doc_events.item.get_diamond_range_rate",
  1950:				response = requests.post(
  1951:					url=_DIAMOND_RATE_URL,
```

the one call in the same file that is bounded:

```python
  935:				response = requests.post(
  ...
  939:					timeout=30,
```

```mermaid
flowchart TD
  A["User saves a Sales Order"] --> B["before_validate hook"]
  B --> C["create_new_bom1"]
  C --> D["_process_single_row, once per item"]
  D --> E["_process_gemstone_detail, once per gemstone row"]
  D --> F["_process_diamond_detail, once per diamond row"]
  D --> G["_get_finding_sub_info"]
  E --> H["requests.post with no timeout"]
  F --> H
  G --> H
  H --> I["remote stall pins the worker, save transaction stays open"]
  I --> J["worker pool exhausted, site-wide outage"]
```

**Fix** — Can you confirm whether bounding these is in scope for this branch? The PR reads as "sales_order.py has been remediated" while its largest availability hole is untouched, and since every one of these lines is already being rewritten for formatting it is the cheapest place it will ever be. Minimum is a module-level `TIMEOUT` passed to all five plus `except requests.exceptions.RequestException` so a remote failure degrades the pricing lane instead of aborting the save; the repo already has the right shape in `customization/quotation/doc_events/remote_po.py` (short `TIMEOUT`, Redis circuit breaker, failures can never raise into a save). If it is deliberately out of scope, saying so in the PR description is enough.

---

### 5. 5 of 7 `test_qc_lifecycle` tests self-skip on CI's fresh site · **MEDIUM**

**What** — `setUpClass` selects an arbitrary Manufacturing Operation that already exists on the site, and `_make_qc` calls `self.skipTest` when that query is empty. CI builds the site from scratch, so `tabManufacturing Operation` is empty and every DB-backed test in the module skips. The two that survive are source-text and constant assertions.

**Why** — `install.sh` runs `bench new-site test_site`, then only installs apps and copies fixtures; nothing creates a Manufacturing Operation, `hooks.py` declares no `before_tests`, the app ships no test records, and `run-tests --module` loads none for unrelated DocTypes. The module is also placed **first** in `manufacturing-tests.yml`, before the `--doctype "Manufacturing Operation"` invocation that might otherwise have left rows behind, so the skip is deterministic. unittest reports skips as success.

**How** — A regression is reintroduced in `QC.force_approve` or `QC._apply_qc_outcome` — the `duplicate_qc` write reverting to `self.save()`, say, or the visited set being dropped. The five tests that would catch it skip, the job is green, the defect merges. The docstring documents the `RecursionError` as "reproduced, not assumed", but the reproduction only ever runs on a developer's populated site.

**When** — Every CI run, every PR. Anyone reading the green Manufacturing Tests check believes the QC lifecycle is guarded.

**Where** — `jewellery_erpnext/jewellery_erpnext/tests/test_qc_lifecycle.py:47`

```python
	@classmethod
	def setUpClass(cls):
		super().setUpClass()
		cls.mop = frappe.db.sql(
			"""
			SELECT name, manufacturing_work_order
			FROM `tabManufacturing Operation`
			WHERE IFNULL(manufacturing_work_order, '') <> ''
			LIMIT 1
			""",
			as_dict=True,
		)

	def _make_qc(self, status):
		if not self.mop:
			self.skipTest("no Manufacturing Operation with a work order on this site")
```

```mermaid
sequenceDiagram
    participant CI as GitHub runner
    participant Site as test_site
    participant Mod as test_qc_lifecycle
    CI->>Site: bench new-site, install apps, copy fixtures
    CI->>Mod: run the registered module
    Mod->>Site: select one Manufacturing Operation with a work order
    Site-->>Mod: no rows
    Mod-->>CI: 5 of 7 skipped, 2 source assertions pass
    CI-->>CI: Manufacturing Tests reports green
```

**Fix** — Build the fixture instead of querying for one: create a minimal Manufacturing Work Order plus Manufacturing Operation in `setUpClass` (the registered `test_mop_weight_recalc` and `test_make_receive_entry_mop_cap` modules already do this on the same fresh site) and delete the `skipTest` branch. If a self-built operation is genuinely impossible, make the absence a failure — `self.fail("fixture missing")` — so a vacuous run cannot report green. Keeping `super().setUpClass()` is correct; only the data source needs to change.

---

### 6. Permission guards are asserted by source-text substring only · **MEDIUM**

**What** — Every assertion that a guard exists is a string search over `inspect.getsource`: `assertIn("frappe.has_permission(", source)` for the four `job_card` endpoints, `assertIn("has_permission", source)` for the `utils` replacement, `assertIn("check_permission", source)` for `add_hold_comment`. The one behavioural test (`:106-112`) asserts `frappe.has_permission("Stock Entry", "create")` directly — it never calls `create_stock_entry`, `create_stock_entry_material_receipt`, `create_internal_transfer` or `create_new_job_card`.

**Why** — A substring match is satisfied by a comment, a docstring, a dead branch, or a call whose result is discarded — `frappe.has_permission(...)` without `throw=True` returns False and continues. `test_guard_precedes_document_creation` only compares `source.find("frappe.has_permission")` against `source.find("frappe.new_doc")`, which a commented-out reference also satisfies. The suite pins the shape of the code, not the behaviour of the endpoint, and it is why issue 2's wrong DocType passes unnoticed.

**How** — A later refactor keeps the guard text but drops `throw=True`, or moves the guard into a branch behind a flag. All six assertions stay green and the endpoints are open again. `test_stock_reconciliation_override` has the same property: it would pass if `super().remove_items_with_no_change()` were called and its result discarded.

**When** — Any subsequent edit to `job_card.py`, `utils.py`, `parent_manufacturing_order.py` or the Stock Reconciliation override. Also today: nothing proves the new guards refuse a real caller.

**Where** — `jewellery_erpnext/jewellery_erpnext/tests/test_job_card_endpoint_permissions.py:49`

```python
	def test_every_stock_creating_endpoint_checks_permission(self):
		"""Each endpoint must assert Stock Entry create rights before doing anything."""
		for name in GUARDED_ENDPOINTS:
			source = inspect.getsource(getattr(job_card, name))
			self.assertIn(
				"frappe.has_permission(",
				source,
				f"{name} creates or submits a document without an authorisation check",
			)
```

**Fix** — Keep the AST sweeps (`test_no_ungated_whitelisted_writer_in_this_module`, `test_no_unscoped_doctype_docname_endpoint`); they generalise the defect class and are the strongest part of the suite. Add one true negative per endpoint inside the existing role-less-user block: `set_user` to the role-less address the test already creates, then `with self.assertRaises(frappe.PermissionError): job_card.create_stock_entry(...)` for each of the four, and assert `frappe.db.count("Stock Entry")` is unchanged. Likewise make `test_stock_reconciliation_override` submit one Stock Reconciliation and assert `difference_amount != 0`, since that number is the whole point of un-stubbing the method.

---

### 7. `create_new_job_card` gates on Job Card create but mutates a submitted Work Order · **LOW**

**What** — The gate is `frappe.has_permission("Job Card", "create")`. When the requested operation is absent from the Work Order, the endpoint appends a `Work Order Operation` row to an already-submitted Work Order and saves it with both `flags.ignore_validate_update_after_submit = True` and `ignore_permissions=True`. The comment directly above the gate identifies this as "a wider hole than the three sibling endpoints" and then gates the narrower half.

**Why** — Job Card create is a shop-floor permission; extending a submitted Work Order's routing is a manufacturing-control one. `ignore_validate_update_after_submit` disables the framework's only protection against editing non-`allow_on_submit` fields after submit, `ignore_permissions=True` disables the role check on that save, and `work_order` is a free-form client argument with no company or ownership scoping. As with issue 2, base had no check at all here, so this is an **incomplete new guard**, not a hole the PR opens.

**How** — A caller with Job Card create supplies any submitted Work Order name and an operation not already on it; the routing that downstream Job Cards and operation costing key on is permanently altered by someone who cannot edit that document through any supported path.

**When** — Any authenticated session with Job Card create, whenever the supplied operation is absent — which is the normal trigger for this endpoint, not an edge case.

**Where** — `jewellery_erpnext/jewellery_erpnext/doc_events/job_card.py:151`

```python
	# Creates a Job Card AND appends a row to an already-submitted Work Order with
	# `ignore_validate_update_after_submit` plus `ignore_permissions` — the permission
	# check and the update-after-submit validation are both disabled on that path, so
	# this is a wider hole than the three sibling endpoints below, not a narrower one.
	# Gate on Job Card create: that is the document this endpoint exists to produce.
	frappe.has_permission("Job Card", "create", throw=True)
...
		work_order_doc.flags.ignore_validate_update_after_submit = True
		work_order_doc.save(ignore_permissions=True)
```

**Fix** — Gate the Work Order mutation where it happens, on the branch that performs it:

```diff
+		frappe.has_permission("Work Order", "write", doc=work_order_doc, throw=True)
 		work_order_doc.flags.ignore_validate_update_after_submit = True
 		work_order_doc.save(ignore_permissions=True)
```

and assert the Work Order's company matches the `company` argument so the endpoint cannot reach across companies. If the intent really is that Job-Card creators may extend submitted routings, say so in the comment and note who signed off — right now the comment flags the hole and the code leaves it open.

---

### 8. `force_approve` terminates a `duplicate_qc` cycle silently, contrary to the comment · **LOW**

**What** — The comment and the commit message say the bound exists so "a cycle or a pathological lineage surfaces as an error instead of recursing until the stack blows". A genuine cycle does not surface as an error: the `while` guard `current.name not in seen` exits cleanly and every document in the cycle is then force-approved. Only a chain of more than 20 *distinct* documents reaches the throw.

**Why** — For A → B → C → A, `seen` grows to three entries and the loop terminates on the fourth iteration because `current.name in seen`. `len(seen)` never exceeds 3, so `len(seen) > self.MAX_FORCE_APPROVE_CHAIN` never fires. The visited set guards termination (correctly), the counter guards chain length; neither reports a cycle. The behaviour is safe — nothing recurses and nothing is corrupted — but it is not what the code says it does.

**When** — Whenever a `duplicate_qc` cycle exists, which is the exact data defect this rewrite was written to handle. `test_qc_lifecycle.py` asserts cycle *termination*, which passes, not cycle *reporting*.

**Where** — `jewellery_erpnext/jewellery_erpnext/doctype/qc/qc.py:132`

```python
		# A rejected QC spawns a replacement, which may itself be rejected and spawn
		# another. Force-approving walks that chain; this bounds it so a cycle or a
		# pathological lineage surfaces as an error instead of recursing until the
		# stack blows.
		MAX_FORCE_APPROVE_CHAIN = 20
...
			while current is not None and current.name not in seen:
				seen.add(current.name)
				chain.append(current)
				if len(seen) > self.MAX_FORCE_APPROVE_CHAIN:
					frappe.throw(
```

**Fix** — Detect the revisit explicitly instead of folding it into the loop condition:

```diff
-		while current is not None and current.name not in seen:
+		while current is not None:
+			if current.name in seen:
+				frappe.throw(
+					_("QC replacement chain from {0} contains a cycle at {1}.").format(
+						frappe.bold(self.name), frappe.bold(current.name)
+					),
+					title=_("QC Chain Cycle"),
+				)
 			seen.add(current.name)
```

If swallowing cycles is the intended behaviour, drop the "surfaces as an error" wording and add an assertion pinning whichever semantics you choose.

---

### 9. The SQL guard only detects quote-adjacent interpolation · **LOW**

**What** — `_interpolates_a_value` flags a placeholder only when the character immediately before it **and** the character immediately after it are both quotes. Two real shapes slip through: an interpolation in an unquoted value position (`LIMIT {n}`, `WHERE qty > {x}`, `OFFSET {p}`), and two adjacent placeholders inside one quoted literal. The module docstring at `:14` nonetheless says "All 9 are now parameterised, and this test fails if a tenth appears".

**Why** — The sentinel is one character per placeholder, so for `f"... '{a}{b}'"` the lookahead after the first sentinel is the second sentinel and the check returns False; for an unquoted position neither neighbour is a quote. MariaDB accepts unquoted numeric literals, so `WHERE qty > {v}` is a complete injection the guard passes. The second assertion, `test_the_nine_fixed_sites_stay_parameterised`, has the same blind spot — its regex also requires quotes. The directory-level `ALLOWLIST` compounds it: `"jewellery_erpnext/patches"` at `:37` exempts the whole patches tree.

**How** — A later PR adds `frappe.db.sql(f"SELECT ... ORDER BY {order_by} LIMIT {limit}")` with both arriving from a whitelisted endpoint's kwargs. The guard passes, the check stays green, and the app gets its first live injection against a baseline documented as having none. The shape already exists in the tree in benign form — `_bulk_update_child_rows` at `sales_order.py:2700-2712` splices `set_clause` and `name_list` into an `UPDATE` with no surrounding quotes, and is safe only because every fragment goes through `frappe.db.escape`, which is an accident of that function rather than something the guard verifies.

**When** — The next f-string SQL in a numeric or identifier position, or anywhere under the allowlisted prefixes.

**Where** — `jewellery_erpnext/jewellery_erpnext/tests/test_sql_parameterisation.py:72`

```python
def _interpolates_a_value(node: ast.JoinedStr) -> bool:
	"""True when a placeholder sits between quote characters, i.e. a value position."""
	template = "".join(
		part.value
		if isinstance(part, ast.Constant) and isinstance(part.value, str)
		else "\x00"
		for part in node.values
	)
	for match in re.finditer("\x00", template):
		i = match.start()
		if template[max(0, i - 1) : i] in ("'", '"') and template[i + 1 : i + 2] in (
			"'",
			'"',
		):
			return True
	return False
```

**Fix** — Treat any `ast.JoinedStr` passed as the first argument to `db.sql` as suspect unless every `FormattedValue` provably resolves to something structural — a module-level name or constant, or a `", ".join(...)` of one. The existing 32 structural sites all satisfy that rule, so it is cheap. Narrow the `ALLOWLIST` from directory prefixes to explicit `path:lineno` entries so a new site under `patches/` still gets reviewed, and soften the docstring to state what the detector actually covers.

---

### 10. Swallowed `EmptyStockReconciliationItemsError` and the message log · **LOW**

**What** — In the MWO / auto lane, `super()` is called inside a `try` and `EmptyStockReconciliationItemsError` is caught with a bare `pass`. Can you confirm the user does not still see the red error on a save that succeeded? `frappe.throw` is `msgprint` with `raise_exception` set, and `msgprint` appends the message to `frappe.message_log` **before** raising — so catching the exception looks like it would leave the message in the response buffer. The partial-trim case would have the mirror problem: upstream's "Removed items with no change in quantity or value." `msgprint` fires, then the `finally` block puts the rows back, so the user is told rows were removed when they were not.

**Where** — `jewellery_erpnext/jewellery_erpnext/customization/stock_reconciliation/stock_reonciliation.py:189`

```python
		original_items = list(self.items)
		try:
			super().remove_items_with_no_change()
		except EmptyStockReconciliationItemsError:
			pass
		finally:
			if len(self.items) != len(original_items):
				self.items = original_items
```

**Fix** — If it does surface, capture `len(frappe.message_log)` before the `super()` call and truncate `frappe.local.message_log` back to it in both the `except` and the rows-restored branch — or pre-check "all rows unchanged" and skip the `super()` call in these lanes rather than relying on catch-and-continue.

---

### 11. The new `custom_auto_creation` branch may guard a lane nothing can enter · **LOW**

**What** — `keep_all_rows` adds a second protected lane, but can you confirm anything still produces a Stock Reconciliation carrying that flag? The only writer is `create_stock_reconciliation` in `stock_reconciliation_template_utils.py:59-71`, and its caller chain looks dead: the only reference, `StockReconciliationtemplate.before_save`, is commented out, and `hooks.py` has no `scheduler_events` entry for it.

**Why** — To be fair to the branch: during that function's own `insert()` the flag *is* 1 in memory, so the lane is not inert at insert time — `db_set("custom_auto_creation", 0)` only zeroes it afterwards, which means no *later* save can see it. The question is purely whether `stock_reconciliation()` still has a live caller. If it does not, the disjunct and the docstring at `:176-179` describe a lane that no longer exists, and the next person reasoning about the trimming behaviour will reason about it.

**Where** — `jewellery_erpnext/jewellery_erpnext/doctype/stock_reconciliation_template/stock_reconciliation_template.py:7-10`

```python
class StockReconciliationtemplate(Document):
	# def before_save(self):
	# 	stock_reconciliation(self)
	pass
```

```python
 59: def create_stock_reconciliation(stock):
 66: 			"custom_auto_creation": 1,
 70: 	stock_reconciliation_doc.insert(ignore_mandatory=True)
 71: 	stock_reconciliation_doc.db_set("custom_auto_creation", 0)
```

**Fix** — Either drop the `custom_auto_creation` disjunct and the flag write, or restore a real caller and stop zeroing the flag before the document is validated again. Whichever it is, make the docstring say so — it currently presents the lane as live. (The field itself does exist as a column, so this is dead wiring rather than a missing field.)

---

### 12. Three assertions pin the source text of frappe internals · **LOW**

**What** — `test_submit_inherits_the_insert_bypass` asserts that `inspect.getsource(frappe.model.document.Document._submit)` contains the literal `"self.save()"`, and that `Document._save`'s source contains `"if ignore_permissions is not None:"`.

**Why** — These are assertions about another repository's formatting. Both strings are present in the installed frappe today (`frappe/model/document.py:1327-1330` and `:573-574`), so they pass — but `return self.save()` becoming `return self._save_with(...)`, or the guard being rewritten as a ternary, turns a `jewellery_erpnext` PR check red for a reason unrelated to that PR. Whoever triages it will have no context and will most likely delete the assertion, taking the staged-removal reasoning with it. The reasoning is valuable; the enforcement mechanism is the wrong one.

**Where** — `jewellery_erpnext/jewellery_erpnext/tests/test_job_card_endpoint_permissions.py:124`

```python
		source = inspect.getsource(frappe.model.document.Document._submit)
		self.assertIn(
			"self.save()",
			source,
			"_submit no longer delegates to a bare save(); re-verify the flag-leak "
			"reasoning behind the staged ignore_permissions removal",
		)

		save_source = inspect.getsource(frappe.model.document.Document._save)
		self.assertIn(
			"if ignore_permissions is not None:",
			save_source,
			"_save no longer guards the flag assignment; the bypass may no longer persist",
		)
```

**Fix** — Replace both with one behavioural assertion, which is shorter and refactor-proof: insert a document with `ignore_permissions=True`, submit it as a role-less user, assert the submit succeeded — that is the flag leak stated as behaviour. Keep the docstring exactly as written; it is the most useful part. The QC and Stock Reconciliation `getsource` assertions can stay, since those pin this repo's own code rather than a dependency's.

---

### 13. Redundant per-row `frappe.db.get_value` inside the Stock Entry item loop · **LOW**

**What** — `make_mop_stock_entry` gains a fallback inside `for row in new_se_doc.items:` that re-reads `Material Request Item.warehouse` one row at a time whenever the bulk map has no truthy value, plus fourteen lines of commented-out SQL in the same loop body.

**Why** — `mri_warehouse_map` is called without `mr`, so its single `frappe.get_all` over `{"name": ("in", sorted(missing))}` already covers every name on the Stock Entry. A falsy entry afterwards means the row is gone or its warehouse is genuinely blank — and `frappe.db.get_value` returns exactly the same. Both bypass permissions, so there is no visibility difference either. The result is a round trip that can never change the outcome, inside a path that ends in `save()` + `submit()`. It is also bundled into a commit labelled "sec: parameterise the 9 SQL sites", where a behaviour-shaped change will not be looked for.

**When** — Every `make_mop_stock_entry` call where at least one Material Request Item has a blank warehouse. Bounded — one extra indexed single-row SELECT per affected row, no correctness risk.

**Where** — `jewellery_erpnext/jewellery_erpnext/customization/material_request/material_request.py:71`

```python
		warehouse_data = mri_warehouse_map(new_se_doc.items)

		for row in new_se_doc.items:
			if not warehouse_data.get(row.material_request_item):
				warehouse_data[row.material_request_item] = frappe.db.get_value(
					"Material Request Item", row.material_request_item, "warehouse"
				)
			s_warehouse = warehouse_data.get(row.material_request_item)
```

base and current upstream tip both read simply:

```python
		for row in new_se_doc.items:
			row.s_warehouse = warehouse_data.get(row.material_request_item)
```

**Fix** — Revert the hunk and delete the commented-out SQL block. If the intent was actually to change behaviour — to fall back to something other than the item's own warehouse when it is blank — put it in its own commit outside the SQL series and cover it in `test_material_request_mop_department`, which is already registered in `manufacturing-tests.yml` and therefore actually runs.

---

## Tests & CI

**Registration: all six new modules run.** This is the criterion most PRs in this batch miss, and this one gets it right —

| Module | Registered in |
|---|---|
| `test_gemstone_size_parser.py` | `core-tests.yml` |
| `test_sql_parameterisation.py` | `core-tests.yml` |
| `test_universal_primitives.py` | `core-tests.yml` |
| `test_stock_reconciliation_override.py` | `inventory-tests.yml` |
| `test_qc_lifecycle.py` | `manufacturing-tests.yml` |
| `test_job_card_endpoint_permissions.py` | `manufacturing-tests.yml` |

**Rollback hygiene is clean.** No new class overrides `setUpClass` with a bare `pass` — `test_qc_lifecycle` correctly calls `super().setUpClass()` — and there is no `frappe.db.commit()` anywhere in the new tests.

**But registered is not the same as covering.** Three of the six modules assert almost nothing about behaviour:

- `test_qc_lifecycle` — 5 of 7 tests skip on CI's fresh site (issue 5). The two that run are source-text and constant assertions.
- `test_stock_reconciliation_override` — a `UnitTestCase` whose six tests are all `get_controller` / `__mro__` / `inspect.getsource`. It never saves a Stock Reconciliation, so it cannot see the behaviour change in issue 1.
- `test_job_card_endpoint_permissions` — substring assertions over source (issue 6), plus three assertions pinned to frappe internals (issue 12).

The AST sweeps in `test_universal_primitives` and `test_sql_parameterisation` are the strongest part of the suite — they generalise the defect class instead of pinning known instances — but both have detector blind spots (issues 3 and 9) that let a real gap through today.

---

| File | Also touched by |
|---|---|
| `jewellery_erpnext/hooks.py` | #660 #1043 #1115 #1201 #1251 #1260 #1272 #1275 #1276 |
| `.github/workflows/core-tests.yml` | #1251 #1260 #1275 #1276 |
| `.github/workflows/inventory-tests.yml` | #1115 #1233 #1260 #1275 #1276 |
| `.github/workflows/manufacturing-tests.yml` | #1233 #1260 #1275 #1276 |

```mermaid
flowchart LR
  A["PR 1045"] --> B["hooks.py"]
  A --> C["core-tests.yml"]
  A --> D["inventory-tests.yml"]
  A --> E["manufacturing-tests.yml"]
  F["PR 1275 and PR 1276"] --> B
  F --> C
  F --> D
  F --> E
  G["PR 1233"] --> D
  G --> E
  H["PR 1251"] --> C
  I["PR 1260 branch sync"] --> B
```

Two notes on ordering:

- The workflow collisions are textual, not semantic — every PR appends to the same `run:` block, so expect conflicts but trivial resolutions. The `hooks.py` collision is different: #1045 rewrites the `override_doctype_class["Stock Reconciliation"]` entry from a two-element list to a single path (`hooks.py:298-306` at head). Anyone else editing that dict needs to know the list form is now deliberately forbidden, or the dead-entry bug comes straight back.
- The cross-PR sweep currently puts #1045 in the "resolve or close" bucket alongside #1043 and #1260. Splitting it as below moves five of the eight commits out of that bucket immediately.

---

## Suggested order of work

1. **Land `7b85845f` on its own, first.** Format-only and AST-verified; landing it separately is what makes the other seven commits reviewable and keeps `sales_order.py` / `sales_invoice.py` out of everyone else's conflict path for one merge only.
2. **`1638f4fe` (gemstone `eval` removal + parser tests) and `82517d7c` (CI registration)** — mergeable as-is.
3. **`178e2ed8` (SQL parameterisation)** — mergeable once the guard docstring matches the detector, or the detector is widened (issue 9).
4. **`6d9d63f6`** — add the `scan_scrap_qr_action` gate and widen the mutation detector (issue 3), then merge. The `db_get_value` removal and the PMO comment gate are good as they stand.
5. **`fd7059a6`** — fix the DocType named by `create_internal_transfer` and the copied comment (issue 2), gate the Work Order mutation in `create_new_job_card` (issue 7), and add the four negative tests (issue 6).
6. **`6a0020b3`** — needs an explicit scope decision on ordinary reconciliations (issue 1) plus one `IntegrationTestCase` that actually saves a document; resolve issues 10 and 11 in the same pass.
7. **`a42a001b`** — decide whether cycles are reported or swallowed and make the comment match (issue 8); build the QC fixture so the tests stop skipping (issue 5).
8. **Separately, and small:** the five `requests.post` timeouts (issue 4). Roughly fifteen lines, and the single highest-value change in the vicinity of this branch.

Happy to go deeper on any of these — particularly the Stock Reconciliation scope call, which is the one that needs a product decision rather than a code change.

PR #1059, opened 2026-08-07
feat: implement robust EOD sync lock management and error handling

DHINESH00:


DHINESH00:
## Review — #1059 Robust EOD sync lock management and error handling

**Verdict:** Merge after fixes
&nbsp;·&nbsp; 0 blocker · 2 high · 0 medium · 4 low · 2 nit
&nbsp;·&nbsp; base `v16_develop_aerele` · head `430dc8d0` · 4 files, +604/-58

---

## What this PR does

Four things, all on the EOD sync failure path:

1. `ensure_eod_sync_lock_released()` is added to `eod_lock.py` and called from `sync_mop_logs`' `finally`, so a run that dies *inside its own failure handler* still clears `eod_sync_running`.
2. `_is_connection_lost` / `_db_exception_classes` / `_reconnect_if_connection_lost` are added to `mop_eod_sync.py`, and a dropped DB connection is folded into the existing "cut short, not broken" recoverable bucket.
3. The top-level `except` is reordered so the lock is released **before** the consolidated Error Log is written, and every remaining diagnostic write is wrapped in its own `try/except`.
4. The scheduler's once-per-day guard is relaxed so a cut-short run earns one same-day retry, capped at two attempts.

Several things here are done unusually well and worth calling out. The driver-class net in `_db_exception_classes` is genuinely the right approach for this bench — `frappe/database/__init__.py:78` really does switch on `use_mysqlclient` and both `MySQLdb` and `pymysql` are installed, so an `isinstance` against only one would go blind. Matching connection loss by MySQL client error **code** (2002/2003/2006/2013) rather than by `OperationalError` class is exactly the right precision, and the tests assert the negative cases (1054/1205/1146, plus a non-DB exception carrying the code) rather than only the happy ones. Releasing the lock before diagnostics is the correct priority and `test_lock_is_released_before_the_error_log_is_written` asserts the *ordering*, not just the calls. `ensure_eod_sync_lock_released` deliberately does not stamp `eod_sync_last_completed_on`, and there is a test for that. F18's deliberate `if sync_log_name:` nesting was correctly left alone. The comments in the diff are tied to dated incidents and every new `frappe.logger()` call site names what to check next — that is the kind of diff that is still readable in a year.

```mermaid
flowchart TD
  A["sync_mop_logs"] --> B["set_eod_sync_running\ncommits eod_sync_running=1"]
  B --> C["plan and commit MWO buckets"]
  C --> D{"exception reaches the\ntop-level handler"}
  D -- "no" --> E["clean path\nrelease_eod_sync_lock"]
  D -- "yes" --> F["NEW _reconnect_if_connection_lost\nswaps a dead DB connection"]
  F --> G["NEW release the lock FIRST\n_release_eod_sync_lock_safely"]
  G --> H["NEW _create_consolidated_error_log\nwrapped in try/except"]
  H --> I["finalize the MOP EOD Sync Log"]
  E --> J["NEW finally\nensure_eod_sync_lock_released"]
  I --> J
  J --> K["NEW scheduler guard\n_scheduler_already_attempted_today"]
  K --> L["one same-day retry for a\ncut-short run, capped at two"]
```

## Before → after

```mermaid
flowchart TD
  subgraph BEFORE["Base — diagnostics first"]
    B1["exception reaches the handler"] --> B2["_create_consolidated_error_log\nwrites tabError Log"]
    B2 -- "write fails on crashed MyISAM table" --> B3["second exception escapes"]
    B3 --> B4["eod_sync_running stays 1\nuntil the 4-hour window expires"]
    B2 -- "write succeeds" --> B5["release_eod_sync_lock"]
  end
  subgraph AFTER["Head — lock first"]
    A1["exception reaches the handler"] --> A2["_reconnect_if_connection_lost"]
    A2 --> A3["_release_eod_sync_lock_safely\neod_sync_running=0"]
    A3 --> A4["_create_consolidated_error_log\nfailure is swallowed and logged"]
    A4 --> A5["finally — ensure_eod_sync_lock_released"]
  end
```

That reordering is a real fix, and the `finally` net closes the remaining hole. The findings below are all about the *fourth* change — the scheduler retry — plus some operator-facing edges of the third.

---

## Issues at a glance

| 1 | **HIGH** | `"Failed"` is treated as retryable, but it is also what a deterministic failure writes — second run produces a second draft Stock Entry for the same material | `scheduler.py:38` | Every bucket fails at submit; `processed_mwos == 0` |
| 2 | **HIGH** | Same-day retry can commit `eod_sync_status="Queued"` for a job the dedupe silently drops, wedging the scheduler permanently | `scheduler.py:161` | Hourly reaper stamps "Timeout Released" while RQ still reports the job STARTED |
| 3 | **LOW** | Retry set omits `"Partially Completed"`, so an OOM after the first committed chunk still forfeits the day | `scheduler.py:38` | Infrastructure fault interrupts a run mid-way — the common case |
| 4 | **LOW** | Early release passes `error_log_name=None`, so `eod_sync_last_error_log` keeps pointing at an older run | `mop_eod_sync.py:559` | The Error Log write fails — exactly the path this PR hardens |
| 5 | **LOW** | The three helpers that perform the recovery have no direct tests | `mop_eod_sync.py:161` | On any future edit to the recovery path |
| 6 | **LOW** | Two new tests run the real `sync_mop_logs` with class rollback disabled, leaving EOD fields committed on the live singleton | `test_mop_settings.py:6695` | Running the module against a shared dev site |
| 7 | **NIT** | `hooks.py` still says "2-hour window" while `_LOCK_HOURS = 4` | `hooks.py:343` | Triaging a stranded lock from the hooks registration |
| 8 | **NIT** | `release_eod_sync_lock`'s `sync_log_name` is accepted and never used; docstring claims otherwise | `eod_lock.py:119` | On any read or edit of this module |

---

### 1. `"Failed"` conflates a deterministic failure with an infrastructure fault, producing a second draft Stock Entry  ·  **HIGH**

**What** — `_RETRYABLE_SYNC_LOG_STATUSES` contains `"Failed"`, but `"Failed"` is also the status a fully deterministic run writes. `_scheduler_already_attempted_today` therefore re-enqueues the heavy EOD job on the very next per-minute tick after an all-buckets-failed run, re-planning the same still-unsynced MOP Logs and producing a second set of draft Stock Entries for the same material.

**Why** — the docstring says the guard distinguishes "failed deterministically, retrying is pointless" from "the DB server was killed, retrying would work fine". The code never records that distinction: `connection_lost` is computed at `mop_eod_sync.py:512` but only feeds message wording, never the Sync Log status. The clean-path selector writes `"Failed"` whenever `blocking` failures exist and `stats["processed_mwos"] == 0` — and `processed_mwos` is incremented only on a successful submit (`mop_eod_sync.py:2005`), inside the `try` whose handler records the draft. A chunk whose Stock Entry fails ERPNext validation appends a failure dict with **no** `advisory` key (so it counts as blocking), saves a **draft** Stock Entry, and leaves the MOP Logs unsynced.

**How** — see the chain below. `_get_unsynced_mop_groups` filters only on `is_synced=0` / `is_cancelled=0` / the creation window, and the draft-failure branch writes only MOP EOD Sync Log Item child rows, never `MOP Log.is_synced`. So the retry re-plans exactly the same rows. An operator following the run's own instruction — *"Open draft Stock Entry &lt;name&gt;, fix the error, and submit manually"* — now has two drafts for one movement; submitting both is a duplicate Stock Entry and duplicate SRE consumption.

**When** — any scheduled run where all buckets fail at the submit phase, i.e. `blocking` failures with `processed_mwos == 0` (a bad batch, a zero/sub-precision qty, a missing SRE — the ordinary deterministic cases this handler exists for). Also fires on a genuine code defect reaching the top-level handler with `recoverable=False`. Capped at exactly two attempts by `_MAX_SCHEDULER_ATTEMPTS_PER_DAY`, so exactly two drafts. Does **not** fire when any MWO succeeded — that writes "Partially Completed", which is correctly excluded (but see issue 3). The base guard (`frappe.db.exists`, one log = attempted) made a second same-day run impossible, so this path is new.

**Where** — `jewellery_erpnext/jewellery_erpnext/doctype/mop_settings/scheduler.py:38`

```python
# scheduler.py:35-43
# Statuses that mean "the run was cut short by something outside the sync" -- an OOM-killed
# or restarted DB server, an RQ job timeout, a stranded lock the hourly job had to clear.
# Retrying these the same day is worth a second attempt; anything else is not.
_RETRYABLE_SYNC_LOG_STATUSES = ("Timeout Released", "Failed")

# scheduler.py:67-73
	if not logs:
		return False
	if len(logs) >= _MAX_SCHEDULER_ATTEMPTS_PER_DAY:
		return True
	# Block while a run is still in flight, or once one has genuinely settled. Only a
	# cut-short run leaves the door open for the retry.
	return not all(log.status in _RETRYABLE_SYNC_LOG_STATUSES for log in logs)
```

```python
# mop_eod_sync.py:464-470 — the clean path that also writes "Failed"
			final_status = (
				"Completed"
				if not blocking and not stats.get("deadline_stopped")
				else "Partially Completed"
				if stats["processed_mwos"] > 0 or stats.get("deadline_stopped")
				else "Failed"
			)
```

```python
# mop_eod_sync.py:2044-2052 — the draft that gets duplicated; MOP Logs stay unsynced
					"suggested_fix": (
						f"Stock Entry {draft_se_name} is saved as Draft but failed to submit. "
						"Stock Reservation Entries were restored (rolled back). Review the "
						"validation error, fix the underlying issue, and submit the draft "
						"manually. MOP Logs for these MWOs remain unsynced until then."
					),
				}
			)
			stats["draft_ses"].append(draft_se_name)
```

```mermaid
flowchart TD
  T1["19:00 run — every bucket fails\nStock Entry submit validation"] --> T2["record_failure appends a blocking\nentry with no advisory key"]
  T2 --> T3["draft Stock Entry MAT-STE-A saved\nMOP Logs stay is_synced=0"]
  T3 --> T4["processed_mwos == 0\nSync Log status = Failed"]
  T4 --> T5["release_eod_sync_lock\nrunning=0, lock_until=now"]
  T5 --> T6["next per-minute tick"]
  T6 --> T7{"_scheduler_already_attempted_today"}
  T7 -- "Failed is in the retryable tuple" --> T8["second full run enqueued"]
  T8 --> T9["_get_unsynced_mop_groups returns\nthe SAME MOP Logs"]
  T9 --> T10["draft Stock Entry MAT-STE-B\nsame rows — two drafts, one movement"]
```

**Fix** — make the retry decision on the actual cause, not on a status string that conflates two meanings. Either:

- **(a)** stamp the cause on the Sync Log and gate on it — a distinct status (`"Cut Short"`) or a Check field `infrastructure_fault` written from the already-computed `connection_lost` / `recoverable` at `mop_eod_sync.py:512-513` — and set `_RETRYABLE_SYNC_LOG_STATUSES = ("Timeout Released", "Cut Short")`, dropping plain `"Failed"`; or
- **(b)** at minimum, refuse the retry when the previous attempt left drafts: in `_scheduler_already_attempted_today`, also fetch `draft_stock_entries` and return `True` when the prior log has any. Re-planning the same MOP Logs cannot help while a draft already holds those rows.

---

### 2. A same-day retry can commit `eod_sync_status="Queued"` for a job that is never enqueued  ·  **HIGH**

**What** — `set_eod_sync_queued()` commits `eod_sync_status="Queued"` **before** `frappe.enqueue(..., job_id="eod_sync", deduplicate=True)`. When a job with that id is still QUEUED/STARTED, frappe's dedupe returns without enqueuing — leaving `"Queued"` committed with no job that will ever clear it.

**Why** — `check_and_enqueue_eod_sync` then returns early at `scheduler.py:106-107` on every subsequent tick, forever, including on following days. The only self-healing routine, `release_expired_eod_sync_lock`, bails at `eod_lock.py:238-239` because the failed run already set `eod_sync_running=0`. The orphan MOP EOD Sync Log also stays at "Queued" forever, which `mop_eod_sync_log.js` polls. Nothing else writes `eod_sync_status` outside the sync job itself, so only the manual Sync button heals it — and nothing raises an alarm.

**How** — the realistic trigger is the hourly reaper. `release_expired_eod_sync_lock` stamps `eod_sync_status="Timeout Released"` precisely so a still-running or hung worker cannot hold the lock (`eod_lock.py:249-275`). A hard-killed worker leaves the RQ job at STARTED in Redis until `StartedJobRegistry` cleanup, so the very next tick — with `"Timeout Released"` now first in `_RETRYABLE_SYNC_LOG_STATUSES` — re-enqueues into a job RQ still reports as STARTED, and the enqueue is silently dropped. (There is a second, much narrower trigger: the failure handler releases the lock and writes the Sync Log status early while the job is still executing its tail — `_create_consolidated_error_log`, `_flush_sync_log_items()`, then `ensure_eod_sync_lock_released` at `mop_eod_sync.py:647`. That is a sub-second race, since the Sync Log reads "Running" and is therefore non-retryable for almost the whole tail.)

**When** — any day the run is timeout-released while the worker process is still alive — i.e. exactly the OOM/flapping-DB scenario this PR targets. To be precise about what is new: the same ordering exists on base, and a run crossing midnight could already reach it, because the base guard only excluded logs with today's `posting_date`. What this PR newly creates is the **same-day** path — a `"Failed"` / `"Timeout Released"` log no longer blocks the next per-minute tick.

**Where** — `jewellery_erpnext/jewellery_erpnext/doctype/mop_settings/scheduler.py:161`

```python
# scheduler.py:161-170
	set_eod_sync_queued(sync_log_name=sync_log_name)
	frappe.enqueue(
		_SYNC_FUNC,
		queue="long",
		timeout=_LOCK_SECONDS,
		enqueue_after_commit=True,
		job_id="eod_sync",
		deduplicate=True,
		sync_log_name=sync_log_name,
	)

# scheduler.py:106-107
	if row.eod_sync_status == "Queued":
		return
```

```python
# apps/frappe/frappe/utils/background_jobs.py:122-125 — the dedupe only logs
		job = get_job(job_id)
		if job and job.get_status(refresh=False) in (JobStatus.QUEUED, JobStatus.STARTED):
			frappe.logger().error(f"Not queueing job {job.id} because it is in queue already")
			return
```

Two obvious escape hatches do not apply: `get_job` calls `create_job_id(job_id)` itself (`background_jobs.py:674-678`), so the site-namespaced lookup does match; and `enqueue_after_commit` only defers `enqueue_call()`, not the dedupe block, which runs synchronously at the top.

```mermaid
sequenceDiagram
  participant Cron as Per-minute cron
  participant Lock as eod_lock
  participant Settings as MOP Settings
  participant RQ as RQ and redis
  Note over RQ: worker hard-killed, job still STARTED in redis
  Cron->>Lock: release_expired_eod_sync_lock
  Lock->>Settings: running=0, status Timeout Released
  Cron->>Cron: next tick, guard now allows a retry
  Cron->>Settings: set_eod_sync_queued commits status Queued
  Cron->>RQ: enqueue job_id eod_sync with deduplicate
  RQ-->>Cron: dropped silently, only an error line is logged
  Note over Settings: status stays Queued with no job behind it
  Cron->>Settings: every later tick returns early on status Queued
```

**Fix** — do not commit "Queued" state for a job that may not be enqueued. Either:

- **(a)** guard the retry on the job actually being gone — `from frappe.utils.background_jobs import is_job_enqueued` and `if is_job_enqueued("eod_sync"): return` right before the Sync-Log creation block at `scheduler.py:136`; or
- **(b)** move `set_eod_sync_queued(...)` to after a confirmed enqueue — `frappe.enqueue` returns the Job or None, so `job = frappe.enqueue(...); if not job: <mark the Sync Log "Skipped">; return`.

Either way, also make the state self-healing: have `release_expired_eod_sync_lock` (or the top of `check_and_enqueue_eod_sync`) reset `eod_sync_status` away from `"Queued"` when `eod_sync_running=0` and `is_job_enqueued("eod_sync")` is False, so an existing wedge clears itself.

---

### 3. The retry set omits `"Partially Completed"`, so the OOM case it was written for still forfeits the day  ·  **LOW**

**What** — the hardened handler in this same PR writes `"Partially Completed"` for every recoverable failure that had already processed at least one MWO, including the lost-connection case. That status is not in `_RETRYABLE_SYNC_LOG_STATUSES`, so the retry never fires for it.

**Why** — the status is chosen at `mop_eod_sync.py:595-599` as `"Partially Completed" if recoverable and stats["processed_mwos"] > 0 else "Failed"`, and `_is_connection_lost` makes `recoverable` True for a lost connection (consulted at `mop_eod_sync.py:238-239`). An OOM kill at 19:06 that lands after the first bucket committed therefore produces "Partially Completed" → `_scheduler_already_attempted_today` returns True → the day is forfeited, which is the exact outcome the docstring at `scheduler.py:54-60` says it is fixing. Nothing rescues those leftovers on a later day either: `_run_backlog_catchup` returns early unless `enable_eod_backlog_catchup` is on, and the app invariants keep both EOD flags off by default.

To be fair about the grade: this is an **incomplete improvement, not a regression** — base forfeited the day for every failure. And `test_partially_completed_run_does_not_refire` shows the exclusion is deliberate, so this may be a considered trade-off rather than an oversight. It is worth a second look because the chunked-commit design means "some buckets landed" is the *normal* shape of a long EOD run, so the common infrastructure fault is the one that gets no retry.

**When** — every infrastructure fault that interrupts a run after the first committed chunk. The new `test_cut_short_run_retries_once` only exercises the zero-progress variant (`_todays_logs("Timeout Released")`), so CI cannot see the gap.

**Where** — `jewellery_erpnext/jewellery_erpnext/doctype/mop_settings/scheduler.py:38`

```python
# mop_eod_sync.py:595-599
						# Partially Completed, not Failed: a cut-short run that committed
						# chunks really did sync them, and the status is what operators triage on.
						"status": "Partially Completed"
						if recoverable and stats["processed_mwos"] > 0
						else "Failed",
```

```mermaid
flowchart TD
  S["run ends"] --> Q{"blocking failures present"}
  Q -- "none" --> C["Completed"]
  Q -- "yes" --> P{"recoverable AND\nprocessed_mwos above zero"}
  P -- "yes" --> PC["Partially Completed"]
  P -- "no" --> F["Failed"]
  TR["Timeout Released\nstamped by the hourly reaper"] --> R1["retryable — one same-day retry"]
  F --> R1
  PC --> R2["not retryable — day forfeited"]
  C --> R2
```

The diagram makes the mismatch visible: the two boxes that reach "retryable" are the one that can also mean a deterministic failure (issue 1) and the reaper's, while the one that most often means "the DB died mid-run" lands on the wrong side.

**Fix** — the cleanest resolution is the same as issue 1: store the cut-short/recoverable flag on the Sync Log and key the retry on that, which fixes both directions at once. If the statuses must stay as they are, widen to include `"Partially Completed"` **and** narrow `"Failed"` by requiring the run to have been recoverable. Also fix the comment at `scheduler.py:35-37`, which claims these statuses only ever mean an external fault — `else "Failed"` at `:599` is the non-recoverable / code-defect branch. Worth adding a scheduler test for `_todays_logs("Partially Completed")` asserting whichever behaviour you decide is intended.

---

### 4. The early lock release drops `error_log_name`, so MOP Settings can point at a previous run's Error Log  ·  **LOW**

**What** — `_release_eod_sync_lock_safely` is called with only `sync_log_name`, so `release_eod_sync_lock(success=False, error_log_name=None)` writes the message *"EOD sync failed. Check the Last EOD Error Log for details."* without updating `eod_sync_last_error_log`. The compensating write added at `:574-583` is itself guarded by `if error_log_name:` — so on exactly the path this PR hardened (the Error Log write failing) there is no correction, and the field keeps whatever the last successful error-log write stored.

**Why** — `eod_lock.py:132-137` writes the message unconditionally but guards the link behind `if error_log_name:`, and nothing anywhere clears a stale value. `_create_consolidated_error_log` genuinely returns `None` without raising when `frappe.log_error` fails (`mop_eod_sync.py:2221-2232`). Before the reordering, `error_log_name` was always computed first and passed in, so the message and the link always agreed.

**How** — MariaDB is OOM-killed mid-run → handler fires → lock released with `error_log_name=None` and the "check the Last EOD Error Log" message committed → the Error Log write fails against the crashed MyISAM `tabError Log` → `if error_log_name:` is False → MOP Settings shows today's failure message next to last Tuesday's `ERR-XXXX`. An operator clicks through and triages an unrelated traceback.

**When** — any failure where writing the consolidated Error Log fails, on a site that has had at least one earlier EOD failure. Misdirection rather than data loss, but it defeats the operator-facing half of the fix.

**Where** — `jewellery_erpnext/jewellery_erpnext/doctype/mop_settings/mop_eod_sync.py:559`

```python
# mop_eod_sync.py:559
		_release_eod_sync_lock_safely(sync_log_name=sync_log_name)

# mop_eod_sync.py:574-583
		if error_log_name:
			# Re-attach the Error Log to MOP Settings, which the early release could not know.
			try:
				frappe.db.set_value(
					"MOP Settings",
					"MOP Settings",
					"eod_sync_last_error_log",
					error_log_name,
				)
				frappe.db.commit()
```

```python
# eod_lock.py:132-137
	else:
		values[
			"eod_sync_message"
		] = "EOD sync failed. Check the Last EOD Error Log for details."
		if error_log_name:
			values["eod_sync_last_error_log"] = error_log_name
```

**Fix** — make the empty case explicit rather than silent. Either clear the pointer in the release (`values["eod_sync_last_error_log"] = error_log_name or ""`) or drop the `if error_log_name:` guard at `:574` and always write `error_log_name or ""`. When there is no Error Log name, point `eod_sync_message` at the Sync Log instead — the traceback is already stamped there as `last_error`. A test asserting the field is cleared when `_create_consolidated_error_log` raises would lock it in.

---

### 5. The three helpers that actually perform the recovery have no direct tests  ·  **LOW**

**What** — `_db_exception_classes` (`:102`), `_reconnect_if_connection_lost` (`:161`) and `_release_eod_sync_lock_safely` (`:195`) are never referenced by the test module. A grep of `test_mop_settings.py` at PR head for all three names returns zero hits. `_is_connection_lost` and `ensure_eod_sync_lock_released` are both covered well; these three are exercised only incidentally, and only on their success paths.

**Why** — `_reconnect_if_connection_lost` is the highest-risk new code in the PR. It calls `db.connect()`, which replaces `_conn` and `_cursor` wholesale (`apps/frappe/frappe/database/database.py:144-149`) without closing the old connection, and it runs at `mop_eod_sync.py:517` ahead of every recovery write. Its `db is None` and `db.connect()`-raises branches never execute in a test. `_release_eod_sync_lock_safely`'s `except` branch — the one that decides whether the lock stays stuck — is likewise never executed, because both new tests mock `release_eod_sync_lock` to succeed. So the single most important behaviour in the PR ("a failing release must not propagate") is asserted only indirectly.

**How** — a future regression in either function fails no test. `_release_eod_sync_lock_safely` quietly starting to re-raise would put the handler back in the 2026-08-05 state this PR fixes.

**When** — on any future edit to the recovery path. Worth noting the tests themselves **do** run in CI: `.github/workflows/inventory-tests.yml:55` invokes `--doctype "MOP Settings"`, which picks up this module. This is a coverage gap, not an unregistered-test gap.

**Where** — `jewellery_erpnext/jewellery_erpnext/doctype/mop_settings/mop_eod_sync.py:161`

```python
# mop_eod_sync.py:161-194 — untested
	def _reconnect_if_connection_lost(exc):
		...
		if not _is_connection_lost(exc):
			return True
		db = getattr(frappe, "db", None)
		if db is None:
			return False
		try:
			db.connect()
		except Exception:
			...
			return False
```

**Fix** — three small mock-only tests; the existing `TestEnsureEodSyncLockReleased` pattern applies directly.

1. `_reconnect_if_connection_lost` returns True and calls `db.connect()` once for `pymysql.err.OperationalError(2013, ...)`; returns True **without** calling `connect()` for `ValueError("x")`; returns False when `connect()` raises.
2. `_release_eod_sync_lock_safely` returns False and does not propagate when `release_eod_sync_lock` raises.
3. `_db_exception_classes("OperationalError")` contains both driver classes.

One related consistency point while you are in there: `ensure_eod_sync_lock_released` reconnects on a bare `except Exception` around `frappe.db.get_value` (`eod_lock.py:163-173`) rather than gating on `_is_connection_lost` the way `_reconnect_if_connection_lost` does — so a `ProgrammingError` (missing column) also triggers a blind connection swap. Gating it the same way would make the two read alike.

---

### 6. The two new `sync_mop_logs` tests leave EOD state committed on the live MOP Settings singleton  ·  **LOW**

**What** — `test_lock_is_released_even_when_error_logging_fails` and `test_lock_is_released_before_the_error_log_is_written` call the real `sync_mop_logs(...)` without patching `set_eod_sync_running` or `ensure_eod_sync_lock_released`, inside a class whose `setUpClass` is a bare `pass`. Both helpers do `frappe.db.set_value` on the singleton followed by `frappe.db.commit()`, so the writes escape the (already disabled) class rollback.

**Why** — `sync_mop_logs` calls `set_eod_sync_running(...)` at `mop_eod_sync.py:359`, before the `try:` at `:378` and outside every mock these tests install (`_resolve_run_range`, `frappe.logger`, `_get_unsynced_mop_groups`, `release_eod_sync_lock`, `_create_consolidated_error_log` — none intercept it). `IntegrationTestCase` registers its rollback via `addClassCleanup(_rollback_db)` *inside* `setUpClass`, so the bare-`pass` override never registers it; and the explicit commits would escape it regardless.

Credit where it is due, and this narrows the finding considerably: this pattern is **pre-existing** — the same class on base already contains `test_no_sync_log_failure` and `test_top_level_exception` doing exactly this, and on base the committed `eod_sync_running=1` with a 4-hour `eod_sync_lock_until` was never cleared at all, because `release_eod_sync_lock` is mocked. This PR's own `ensure_eod_sync_lock_released` in the `finally` is what now clears it. The PR shrinks the leak while adding two more instances of the pattern.

**How** — what genuinely remains at head: while the test runs, `is_eod_sync_locked()` is True for every other connection on that site, so `validate_not_eod_sync_locked` (wired on `before_save`/`before_submit`/`before_cancel` for Employee IR, Department IR, MOP Log, Stock Entry and Stock Reconciliation at `hooks.py:85-112`, `:178-199`, `:293-295`) throws for anyone else. Afterwards the singleton is left with `eod_sync_status="Failed"`, `eod_sync_lock_until`, `eod_sync_started_on` and the message *"EOD sync ended without releasing its lock…"* — so the MOP Settings screen reports a failure that never happened.

**When** — any developer running `bench --site gk run-tests --doctype "MOP Settings"` against a shared dev site. Not reachable on `kg-gk`: `run-tests` refuses without `allow_tests`, which that site does not have. CI is also unaffected — `.github/workflows/inventory-tests.yml:56` runs against a throwaway `test_site` — which is precisely why CI will never surface this.

**Where** — `jewellery_erpnext/jewellery_erpnext/doctype/mop_settings/test_mop_settings.py:6695`

```python
# test_mop_settings.py:6619-6622 — class rollback disabled
class TestWMopEodSyncErrorHandling(IntegrationTestCase):
	@classmethod
	def setUpClass(cls):
		pass
```

```python
# mop_eod_sync.py:359 — unmocked, outside the try
	set_eod_sync_running(sync_log_name=sync_log_name)
```

**Fix** — add `@patch(f"{_MOD}.set_eod_sync_running")` and `@patch(f"{_MOD}.ensure_eod_sync_lock_released")` to both new tests. Both names are imported into the `mop_eod_sync` namespace, so patching them there is a clean seam and does not weaken what the tests assert — they assert on `release_eod_sync_lock` ordering, not on lock acquisition. Retro-fitting the same two patches to the pre-existing `test_no_sync_log_failure` / `test_top_level_exception` would close the older hole at the same time.

---

### 7. `hooks.py` still advertises a 2-hour lock window  ·  **NIT**

`hooks.py:343-345` reads:

```python
	# Hourly cleanup: auto-releases any EOD sync lock that outlived its 2-hour window.
	"hourly": [
		"jewellery_erpnext.jewellery_erpnext.doctype.mop_settings.eod_lock.release_expired_eod_sync_lock"
```

while `eod_lock.py:23-24` is `_LOCK_HOURS = 4` / `_LOCK_SECONDS = _LOCK_HOURS * 60 * 60`. To be clear, this is not introduced here — `hooks.py` is not in this PR's diff at all, and both the comment and `_LOCK_HOURS = 4` predate it (it is the known F16). Can you confirm you would rather keep `hooks.py` out of scope? If you are happy to take it, a drift-proof wording while the lock story is fresh would be:

```
# Hourly cleanup: auto-releases any EOD sync lock that outlived its window (eod_lock._LOCK_HOURS).
```

The reason it is worth the one-line change: the 2026-08-05 incident this PR is about was a lock stranded for the full window, and the on-call reference says half the real duration.

### 8. `release_eod_sync_lock`'s `sync_log_name` is inert and its docstring says otherwise  ·  **NIT**

```python
# eod_lock.py:115-140 — full body; sync_log_name is never referenced
def release_eod_sync_lock(success=True, error_log_name=None, sync_log_name=None):
	"""Clear running state after EOD sync finishes (success or failure).

	Commits immediately so other connections stop seeing the lock.
	Optionally finalizes the MOP EOD Sync Log.
	"""
```

This is pre-existing (F17) and the file's only change in this PR is the added `ensure_eod_sync_lock_released`, so nothing here regressed — the new `_release_eod_sync_lock_safely` just adds a fourth call site that forwards the inert kwarg. Can you confirm the parameter is meant to stay as a no-op placeholder? The reason to ask rather than ignore it: the Sync Log really is finalized, but by the caller at `mop_eod_sync.py:589-617` and by the new `ensure_eod_sync_lock_released` — so a future reader who trusts the docstring and deletes `:589-617` as redundant would leave the Sync Log at "Running" forever, with `mop_eod_sync_log.js` polling it and `release_expired_eod_sync_lock` unable to rescue it (`eod_lock.py:238-239`). Dropping the parameter from the signature and its call sites, or implementing it the way `ensure_eod_sync_lock_released` already does at `eod_lock.py:197-209`, closes that.

---

## Tests & CI

- **No new test module is added** — `test_mop_settings.py` is modified in place, and it is already reachable in CI: `.github/workflows/inventory-tests.yml:55` invokes `--doctype "MOP Settings"`. So unlike several other open PRs, the new tests here genuinely execute. Good.
- **Coverage gaps** — issue 5: `_db_exception_classes`, `_reconnect_if_connection_lost` and `_release_eod_sync_lock_safely` have zero direct references in the test module. Issue 3: `test_cut_short_run_retries_once` covers only the zero-progress variant, so the "Partially Completed" path is untested in either direction.
- **`setUpClass: pass`** — `TestWMopEodSyncErrorHandling` (`test_mop_settings.py:6619-6622`) disables `IntegrationTestCase`'s class rollback. Pre-existing for the class, extended by two tests here; see issue 6.
- **Commits inside tests** — no `frappe.db.commit()` is written in the test bodies, but two of them reach one indirectly through `set_eod_sync_running` (`eod_lock.py:91`) and `ensure_eod_sync_lock_released`, which is the same thing as far as rollback is concerned.
- Worth adding alongside the fixes: a scheduler test for the `"Queued"`-with-no-job state (issue 2), and one asserting a second same-day run is refused when the prior log left drafts (issue 1).

None. Across the 20 open PRs, the contended files are `patches.txt` (12 PRs), `hooks.py` (10), `manufacturing_operation.py` (7), `doc_events/stock_entry.py` (7), `manufacturing_work_order.py` (6), `create_test_data.py` (6) and the four workflow files — this PR touches none of them. Its four files (`eod_lock.py`, `mop_eod_sync.py`, `scheduler.py`, `test_mop_settings.py`) are not edited by any other open PR, and it is the only PR in the set currently reporting `mergeable_state: clean`. No ordering constraint applies; it can land first and independently.

## Suggested order of work

1. **Decide the retry key** (issues 1 and 3 together). Both come from `"Failed"` and `"Partially Completed"` carrying two meanings each. Storing the already-computed `recoverable` / `connection_lost` flag on the Sync Log and gating on that resolves both; the status-tuple edit alone resolves neither cleanly.
2. **Close the "Queued" wedge** (issue 2) — `is_job_enqueued("eod_sync")` guard, or commit the status only after a confirmed enqueue, plus the self-healing reset so an existing wedge clears.
3. **Clear the stale Error Log pointer** (issue 4) — a one-line change in the release plus a softened message.
4. **Add the three helper tests and the two `@patch` lines** (issues 5 and 6) — both are small and can go in the same commit.
5. The two nits (7, 8) are optional; the `hooks.py` comment is the more useful of the two given the incident it relates to.

Happy to go deeper on any of these — particularly the retry-key design in item 1, where there is a real choice between a new Sync Log status and a Check field.

PR #1103, opened 2026-08-18
Feat/dummy finding process

manikandan-s-18:


Praveenkumar26-S (approved):

DHINESH00:
## Review — #1103 Dummy finding process

&nbsp;·&nbsp; 1 blocker · 1 high · 1 medium · 2 low · 1 nit
&nbsp;·&nbsp; base `kggk_uat` · head `88fb371a` · 6 files, +139/-6

This is already approved, so one note up front: the approval missed a patch that aborts `bench migrate`. I checked the three live site databases — the Item Group the new patch hard-codes exists on `gk` only; `alfarsi` and `kg-gk` do not have it, and neither has the Item, so the `frappe.db.exists` early return will not save them. On those two sites `bench migrate` stops at this patch and the 61 patches after it never run. That is worth a second pass before this lands, and nothing below is a criticism of the review — the failure is invisible from the diff alone and only shows up when you query the sites.

---

## What this PR does

It extends the existing "dummy gemstone item" pattern to findings: a `default_finding_item` Link field on Manufacturing Setting, a patch that creates the `F-PER-DUM-PRE-CC` Item, a `validate_finding_alternative_items` gate in `doc_events/material_request.py` invoked from `create_stock_entry`, and a PMO branch that substitutes the dummy for real finding items when building the Material Request.

Several things are done well and worth saying. The `if/elif/else` rewrite at `parent_manufacturing_order.py:654-659` is strictly clearer than the nested ternary it replaces and keeps `other_item`/`metal_item`/`diamond_item` byte-identical. The patch is genuinely re-runnable (guard plus early return). It is registered in `patches.txt` in the correct `[post_model_sync]` section, immediately after its `default_gemstone_item` sibling. `create_test_data.py` was updated in the same PR so the CI fixture site gets both the Item and the setting. And the new field lives in the app-owned `manufacturing_setting.json`, which migrate schema-syncs, rather than in the inert `custom_fields/` tree — that is the right call and it is the one people get wrong most often in this repo. The concerns below are all about the finding-specific consequences that the gemstone precedent does not cover.

```mermaid
flowchart TD
  A["patches.txt line 16 registers default_finding_item"] --> B["patch creates Item F-PER-DUM-PRE-CC"]
  C["manufacturing_setting.json adds Link field default_finding_item"] --> D["PMO create_material_requests"]
  B --> D
  D --> E{"item_type"}
  E -->|"gemstone_item"| F["substitute default_gemstone_item"]
  E -->|"finding_item"| G["substitute default_finding_item\nor throw when unset"]
  E -->|"anything else"| H["keep the real item_code"]
  F --> I["Material Request rows"]
  G --> I
  H --> I
  I --> J["workflow_state Material Reserved"]
  J --> K["create_stock_entry"]
  K --> L["validate_gemstone_alternative_items"]
  K --> M["validate_finding_alternative_items new"]
  M --> N["Stock Entry line uses custom_alternative_item or item_code"]
```

---

## Before → after

```mermaid
flowchart LR
  B0["Before this PR"] --> B1["finding MR row carries the real F variant item"]
  B1 --> B2["custom_variant_of fetches F"]
  B2 --> B3["custom_pure_qty computed from metal purity"]
  A0["After this PR"] --> A1{"default_finding_item set for this manufacturer"}
  A1 -->|"no"| A2["throw aborts PMO submit and no MWO is created"]
  A1 -->|"yes"| A3["finding MR row carries F-PER-DUM-PRE-CC"]
  A3 --> A4["custom_variant_of is NULL"]
  A4 --> A5["custom_pure_qty stays 0"]
  A3 --> A6["operator picks custom_alternative_item"]
  A6 --> A7["Stock Entry line resolves to the real F variant"]
```

---

## Issues at a glance

| 1 | **BLOCKER** | Patch aborts `bench migrate`, stranding 61 later patches | `jewellery_erpnext/patches/default_finding_item.py:20` | Any site without the `Finding DNU` Item Group — confirmed on `alfarsi` and `kg-gk` |
| 2 | **HIGH** | New throw has no default and no backfill, so PMO submit breaks for every existing manufacturer | `parent_manufacturing_order.py:626` | First PMO with a finding row after migrate, on every site |
| 3 | **MEDIUM** | Dummy substitution zeroes `custom_pure_qty` on every finding MR row | `parent_manufacturing_order.py:657` | Every PMO whose Tracking BOM has a finding row with qty > 0 |
| 4 | **LOW** | New validator and PMO branch ship with zero tests; the gemstone twin has five in CI | `doc_events/material_request.py:94` | Permanently, on every future change to this path |
| 5 | **LOW** | Validator fails open, letting the non-stock dummy reach the Stock Entry | `doc_events/material_request.py:117` | Setting cleared after the MRs were generated |
| 6 | **NIT** | Gemstone and finding validators are a verbatim copy; first throw wins | `doc_events/material_request.py:701` | Maintenance; a mixed MR the PMO does not itself produce |

---

### 1. Patch aborts `bench migrate` when Item Group `Finding DNU` is missing  ·  **BLOCKER**

**What** — The patch inserts an Item with a hard-coded `item_group` of `"Finding DNU"` and no existence guard. Where that Item Group is absent, `item.insert()` raises a link validation error, `bench migrate` aborts at this patch, and the 61 patches listed after it in `patches.txt` never run.

**Why** — `item_group` is `reqd=1` and a Link to Item Group on the ERPNext Item DocType, so a missing target raises during insert. `"Finding DNU"` is created nowhere in the shipped migration path: it exists only in `create_test_data.py`, which is dev/test bootstrap, not a patch and not a fixture. Neither `jewellery_erpnext/fixtures/` nor `gke_customization/fixtures/` contains an `item_group.json`.

**How** — `bench --site <site> migrate` → patch `jewellery_erpnext.patches.default_finding_item` → `item.insert(ignore_permissions=True)` raises for `item_group` → migrate exits non-zero → `patches.txt:17-77` (61 entries, no comments or blanks among them) are left unapplied, while the new `default_finding_item` DocType field is already present because model sync ran first.

**When** — Confirmed against the live databases: Item Group `Finding DNU` exists **only on `gk`**. It is absent on `alfarsi` and on `kg-gk`, the production-refreshed site, and `F-PER-DUM-PRE-CC` does not exist on any of the three, so the `frappe.db.exists` short-circuit at line 11 does not fire. This is the 9th of 70 active `[post_model_sync]` entries, so the abort is close to maximally expensive.

The gemstone sibling is not a precedent that this is safe. On `alfarsi` and `kg-gk`, `G-PER-DUM-PRE-CC` already existed as a real item (`item_group` `Gemstone - V`, `is_stock_item=1`), so `default_gemstone_item` returned early and never exercised the `Gemstone DNU` link. It got lucky. CI cannot catch this either, because `create_test_data.py` creates `Finding DNU` on test sites only.

**Where** — `jewellery_erpnext/patches/default_finding_item.py:20`

```python
	if frappe.db.exists("Item", item_code):
		print(f"Item already exists: {item_code}")
		return

	item = frappe.get_doc(
		{
			"doctype": "Item",
			"item_code": item_code,
			"item_name": "Default Dummy Finding",
			"item_group": "Finding DNU",
			"stock_uom": "Gram",
			"is_stock_item": 0,
			"disabled": 0,
		}
	)

	item.insert(ignore_permissions=True)

	frappe.db.commit()

	print(f"Created Item: {item_code}")
```

The group is created only in test bootstrap — `create_test_data.py:3630-3636`:

```python
	if not frappe.db.exists("Item Group", "Finding DNU"):
				"item_group_name": "Finding DNU",
				"parent_item_group": "DNU",
```

```mermaid
flowchart TD
  M["bench migrate"] --> P1["patch default_gemstone_item"]
  P1 --> P2["patch default_finding_item"]
  P2 --> Q{"Item Group Finding DNU present"}
  Q -->|"yes, gk only"| OK["Item inserted and migrate continues"]
  Q -->|"no, alfarsi and kg-gk"| ERR["link validation error on item_group"]
  ERR --> AB["migrate exits non-zero"]
  AB --> ST["patches.txt lines 17 to 77, 61 patches, never run"]
  AB --> SC["field default_finding_item exists but the Item it points at does not"]
```

**Fix** — Create the group if missing, or at minimum fail soft rather than aborting the whole migrate:

```diff
 	if frappe.db.exists("Item", item_code):
-		print(f"Item already exists: {item_code}")
 		return
 
+	if not frappe.db.exists("Item Group", "Finding DNU"):
+		frappe.log_error(
+			"default_finding_item patch skipped: Item Group 'Finding DNU' is missing",
+			"default_finding_item",
+		)
+		return
+
 	item = frappe.get_doc(
```

Two smaller points in the same file: replace the two `print()` calls with `click.secho` or `frappe.log_error` so the output is captured by bench, and drop the explicit `frappe.db.commit()` — the patch runner commits after each patch, and committing here only makes a partial state durable.

---

### 2. No default and no backfill, so the new throw blocks PMO submit on every existing manufacturer  ·  **HIGH**

**What** — The PR introduces a blocking `frappe.throw` when `Manufacturing Setting.default_finding_item` is empty, but nothing in the PR ever populates that field on an existing site.

**Why** — Three things have to line up and none of them do. `patches/default_finding_item.py` stops at `item.insert()` and never writes `default_finding_item` onto any Manufacturing Setting row. `manufacturing_setting.json:393-398` declares the field with only fieldname/fieldtype/label/options — no `default`, no `reqd`. And `create_test_data.py` sets it only inside `if not frappe.db.exists("Manufacturing Setting", "Shubh")`, which is skipped wherever Shubh already exists. So immediately after `bench migrate` the value is NULL for every manufacturer.

**How** — Worth being precise about the blast radius, because it is wider than "material request creation": `create_material_requests` is not an operator-invoked button. It has no `@frappe.whitelist()` and is called only from `ParentManufacturingOrder.on_submit` at line 404, guarded by `if self.type != "Finding Manufacturing"`. So the throw aborts the whole PMO submission, which also means `create_manufacturing_work_order(self)` at line 405 never runs. And because the throw happens mid-loop over the item types (metal → diamond → finding → gemstone → other) and `mr_doc.save()` never commits, the metal and diamond Material Requests created earlier in the same submit roll back with it. Nothing is created at all.

**When** — Every site, on the first PMO with a finding row after this merges, for every manufacturer, until someone hand-edits `default_finding_item` on each Manufacturing Setting record. I checked the live rows: `alfarsi` has 1 Manufacturing Setting, `gk` has 4, `kg-gk` has 2 — all of them carry only `default_gemstone_item` values that were set by hand after the gemstone PR, and two of the six have even that NULL. No runbook step ships with this PR.

On volume: finding rows are common but not overwhelmingly more common than gemstone rows — on `gk`, 39,067 of 72,544 BOMs have finding rows versus 32,609 with gemstone rows; on `kg-gk` it is 40,809 versus 11,537 of 68,395. The finding stands without needing the "findings are on essentially every BOM" framing.

**Where** — `jewellery_erpnext/jewellery_erpnext/doctype/parent_manufacturing_order/parent_manufacturing_order.py:626`

```python
			if item_type == "finding_item":
				default_finding_item = frappe.db.get_value(
					"Manufacturing Setting",
					{"manufacturer": self.manufacturer},
					"default_finding_item",
				)
				if not default_finding_item:
					frappe.throw(
						_(
							"Default Finding Item is not set in Manufacturing Setting. Please set this item F-PER-DUM-PRE-CC"
						)
					)
```

The field, with no `default` and no `reqd` — `manufacturing_setting.json:393-398`:

```json
  {
   "fieldname": "default_finding_item",
   "fieldtype": "Link",
   "label": "Default Finding Item",
   "options": "Item"
  },
```

And the patch, ending after the Item insert — `patches/default_finding_item.py:27-31`:

```python
	item.insert(ignore_permissions=True)

	frappe.db.commit()

	print(f"Created Item: {item_code}")
```

```mermaid
sequenceDiagram
    participant Op as Operator
    participant PMO as PMO on_submit
    participant MR as create_material_requests
    participant MS as Manufacturing Setting
    Op->>PMO: Submit the PMO
    PMO->>MR: call at line 404
    MR->>MR: metal MR saved
    MR->>MR: diamond MR saved
    MR->>MS: read default_finding_item
    MS-->>MR: None
    MR-->>Op: throw Default Finding Item is not set
    Note over Op,MS: submit aborts, the metal and diamond MRs roll back, MWOs are never created
```

**Fix** — Backfill in the same patch, right after the Item is created, so migrate leaves every site in a working state:

```diff
 	item.insert(ignore_permissions=True)
 
+	# Point every Manufacturing Setting that has no finding placeholder yet at this item,
+	# otherwise PMO submit throws for every existing manufacturer.
+	for name in frappe.get_all(
+		"Manufacturing Setting",
+		filters={"default_finding_item": ("in", ["", None])},
+		pluck="name",
+	):
+		frappe.db.set_value(
+			"Manufacturing Setting", name, "default_finding_item", item_code, update_modified=False
+		)
+
 	frappe.db.commit()
```

Move the backfill above the existing-item early return (or drop that early return and make the backfill the unconditional tail), so a site that already has the Item still gets the link. Worth doing the same for `default_gemstone_item` while you are in there — two of the six live rows are still NULL.

---

### 3. Substituting the dummy zeroes `custom_pure_qty` on every finding MR row  ·  **MEDIUM**

**What** — Findings are metal items with purity attributes. Replacing the real F-variant item code with the non-variant dummy makes the fetched `custom_variant_of` NULL, which switches off the pure-metal calculation for those rows — and it stays off even after the operator picks the real item in `custom_alternative_item`. Gemstones were immune because variant `G` is not in the `M`/`F` gate; findings are not.

**Why** — `custom_variant_of` is `fetch_from: item_code.variant_of`, derived from `item_code` and never from `custom_alternative_item`. The patch creates `F-PER-DUM-PRE-CC` with no `variant_of` key, so the fetch yields NULL. `_is_pure_qty_row` gates purely on `row.custom_variant_of in ["M", "F"]`, so the row drops out of both the prefetch pass and the assignment loop. The fetch is not rescuable client-side either: the live Custom Field row on both `gk` and `kg-gk` has `fetch_if_empty=0`, so it is overwritten from `item_code` on every save.

Note the inconsistency the PR exposes: inside the loop the code already resolves purity from `row.custom_alternative_item or row.item_code`, so the alternative was clearly meant to drive purity — but the *row gate* still reads the dummy's own variant.

**How** — PMO appends a finding row with `item_code = F-PER-DUM-PRE-CC` → `mr_doc.save()` → `before_validate` → `update_pure_qty` → `custom_variant_of` is NULL → `_is_pure_qty_row` returns False → `custom_pure_qty` stays 0 while `custom_total_quantity` still accumulates `row.qty`. The operator then sets a real alternative and saves again, and the gate still reads the dummy, so it is still 0. `update_pure_qty` runs on every save via `hooks.py:275` → `material_request.py:172`.

**When** — Every PMO whose Tracking BOM has a `BOM Finding Detail` row with qty > 0 and `custom_transfer_type != "Transfer To Branch"`. For scale: `tabMaterial Request Item` today holds 29,850 rows with `custom_variant_of='F'` on `gk`, 28,708 of which have a non-zero `custom_pure_qty`; on `kg-gk` it is 4,434 and 4,426. All of those would be 0 under this change.

The blast radius is genuinely limited and worth stating so nobody over-reads this: `create_stock_entry` never copies `custom_pure_qty` onto the Stock Entry — the SE recomputes it from its own line's `item_code`, which is `row.custom_alternative_item or row.item_code`, i.e. the real F variant. So no stock, valuation or MOP posting goes wrong. What breaks is the stored and displayed fine-gold figure on a document operators reserve metal against. That is why this is MEDIUM and not HIGH.

**Where** — `jewellery_erpnext/jewellery_erpnext/doctype/parent_manufacturing_order/parent_manufacturing_order.py:657`

```python
					if item_type == "gemstone_item":
						item_code_to_use = default_gemstone_item
					elif item_type == "finding_item":
						item_code_to_use = default_finding_item
					else:
						item_code_to_use = i["item_code"]
```

The gate, in `customization/material_request/utils/before_validate.py:11-20`:

```python
def _is_pure_qty_row(self, row):
	"""Rows whose pure quantity is derived from metal purity.
	return (
		row.custom_variant_of in ["M", "F"]
		and self.custom_transfer_type != "Transfer To Branch"
	)
```

and a few lines below it, `before_validate.py:56-67`, where the purity lookup already keys off the alternative:

```python
			item_purity = get_purity_percentage(
				row.custom_alternative_item or row.item_code
			)
				row.custom_pure_qty = flt((item_purity * row.qty) / pure_item_purity, 3)
```

```mermaid
flowchart TD
  R["MR row item_code is F-PER-DUM-PRE-CC"] --> F["fetch_from item_code.variant_of runs in _validate_links"]
  F --> V["custom_variant_of is NULL"]
  V --> G{"_is_pure_qty_row gate on M or F"}
  G -->|"False"| Z["custom_pure_qty stays 0"]
  ALT["operator sets custom_alternative_item to the real F variant"] --> V2["fetch re-runs from item_code and is still NULL"]
  V2 --> G
  Z --> D["MR shows zero fine gold while the Stock Entry stays correct"]
```

**Fix** — Make the row gate follow the same `alternative or item_code` resolution the purity lookup already uses:

```diff
-def _is_pure_qty_row(self, row):
-	return (
-		row.custom_variant_of in ["M", "F"]
-		and self.custom_transfer_type != "Transfer To Branch"
-	)
+def _is_pure_qty_row(self, row):
+	if self.custom_transfer_type == "Transfer To Branch":
+		return False
+	variant_of = row.custom_variant_of
+	if not variant_of and row.custom_alternative_item:
+		# the row carries a dummy placeholder item; the real metal identity is the alternative
+		variant_of = frappe.get_cached_value("Item", row.custom_alternative_item, "variant_of")
+	return variant_of in ["M", "F"]
```

The cleaner alternative, which also fixes the pre-alternative state: give the dummy a real `variant_of` of the F template in both `patches/default_finding_item.py` and `create_test_data.py`, so `custom_variant_of` fetches `F` from the placeholder itself. Either way, please add a test asserting `custom_pure_qty > 0` on a finding row once the alternative is selected.

---

### 4. The new validator and PMO branch ship with zero tests  ·  **LOW**

**What** — Two new decision branches (`validate_finding_alternative_items` and the PMO `finding_item` substitution) with no test for either, while the gemstone versions of both have tests that do run in CI.

**Why** — `jewellery_erpnext/jewellery_erpnext/tests/test_material_request_customizations.py` is registered at `.github/workflows/manufacturing-tests.yml:60` and its `TestValidateGemstoneAlternativeItems` class (starting at line 34) has five tests: non-reserved state, non-manufacture type, missing alternative, alternative equals dummy, valid alternative. `test_parent_manufacturing_order.py` has `test_create_material_requests_throws_missing_default_gemstone` at line 311 and `test_create_material_requests_uses_default_gemstone` at line 386. The finding path is a line-for-line copy of both and got neither. A grep across the PR head for `validate_finding_alternative_items`, `default_finding_item` and `F-PER-DUM-PRE-CC` returns 21 hits and not one is under a `tests/` directory or in a `test_*.py` — the only non-source hits are the `create_test_data.py` fixture values.

**When** — Immediately and permanently. Any future edit to the finding branch passes CI silently, and because it is a copy rather than a shared helper, the gemstone tests will not catch drift either.

**Where** — `jewellery_erpnext/jewellery_erpnext/doc_events/material_request.py:94`

**Fix** — Add a `TestValidateFindingAlternativeItems` class to `tests/test_material_request_customizations.py`, mirroring the gemstone class. That module is already in the CI allowlist, so **no workflow edit is needed** — the tests will actually run. Plus a `test_create_material_requests_throws_missing_default_finding` in `test_parent_manufacturing_order.py`, and one test asserting `custom_pure_qty` is non-zero on a finding row once the alternative is chosen, which is exactly what would have caught issue 3.

---

### 5. `validate_finding_alternative_items` fails open  ·  **LOW**

**What** — The validator returns silently instead of enforcing whenever `default_finding_item` cannot be resolved. If that happens on an MR whose rows already carry the dummy, the dummy flows into `create_stock_entry` and ERPNext fails at SLE time with a message that names neither the Material Request nor the real cause.

**Why and when** — Narrowed to one reachable trigger: an administrator clears `Manufacturing Setting.default_finding_item` after the MRs were generated, or the MR was hand-built and never got `custom_manufacturer`. The other paths I first suspected do not survive checking — `create_material_requests` sets `mr_doc.custom_manufacturer = self.manufacturer` on every MR it builds, and the throw in issue 2 guarantees the setting was populated at creation time, so the `get_user_default` fallback is never taken on the path this PR creates. Two more reasons this is LOW rather than higher: the same fail-open is a verbatim property of the pre-existing `validate_gemstone_alternative_items` on `kggk_uat`, so the PR copies it rather than introducing it; and the failure is a loud throw that blocks the reservation with nothing written — no corruption, no silently wrong number.

**Where** — `jewellery_erpnext/jewellery_erpnext/doc_events/material_request.py:117`

```python
	manufacturer = self.custom_manufacturer or frappe.defaults.get_user_default(
		"manufacturer"
	)

	if not manufacturer:
		return

	default_finding_item = _get_default_finding_item(manufacturer)

	if not default_finding_item:
		return
```

The dummy then reaches the SE line at `material_request.py:806`:

```python
					"item_code": row.custom_alternative_item or row.item_code,
```

and since the patch sets `"is_stock_item": 0`, `erpnext/stock/doctype/stock_ledger_entry/stock_ledger_entry.py:322-323` throws:

```python
		if item_detail.is_stock_item != 1:
			self.throw_error_message(f"Item {self.item_code} must be a stock Item")
```

**Fix** — Recognise the placeholder structurally rather than depending on the setting at enforcement time: on any Manufacture MR row reaching "Material Reserved", check `frappe.get_cached_value("Item", row.item_code, "is_stock_item")` and throw with the row numbers and the manufacturer name when a non-stock placeholder is still present and no default is configured. Since the gemstone twin has the same hole, fixing it once in a shared helper (see issue 6) closes both.

---

### 6. Gemstone and finding validators are a verbatim copy; first throw wins  ·  **NIT**

`validate_finding_alternative_items` (`material_request.py:94-138`) is character-for-character `validate_gemstone_alternative_items` (`:49-91`) with one word swapped, and `create_stock_entry` calls them back to back:

```python
def create_stock_entry(self, method):
	validate_gemstone_alternative_items(self)
	validate_finding_alternative_items(self)
```

Because each ends in its own `frappe.throw`, an MR holding both an unmapped gemstone row and an unmapped finding row shows only the gemstone errors. In practice the PMO never produces such an MR — it builds one Material Request per item_type — so the user-facing consequence is close to nil; this is a maintainability point. The real cost is that the two copies can drift, and with no tests on the finding side (issue 4) the drift would be invisible.

**Fix** — Collapse them into one parameterised helper that accumulates both error sets and throws once, keeping the two public names as thin wrappers so the existing registered gemstone tests keep passing. That also gives you one place to fix the fail-open in issue 5.

(The `enumerate(self.items, 1)` versus `row.idx` difference is not worth changing — child rows are reindexed in `idx` order, so the two produce the same number on any saved document.)

---

## Tests & CI

- **No test file is added or modified in this PR**, and no `.github/workflows/*-tests.yml` is touched. The mechanical "new test file must be registered" rule therefore does not fire — but the coverage gap is total. None of throw-on-missing-alternative, throw-on-alternative-equals-dummy, pass-on-valid-alternative, or the new PMO `Default Finding Item is not set` throw is exercised by anything.
- A green CI run on this PR proves only that the **gemstone** paths still work.
- The good news is that the fix needs no workflow edit: `jewellery_erpnext.jewellery_erpnext.tests.test_material_request_customizations` is already registered at `manufacturing-tests.yml:60`, so a new test class in that module runs immediately.
- `create_test_data.py` was correctly updated so the CI fixture site gets both the Item and the setting value — that is the right instinct, but it is also why CI cannot see issue 1: the test site has `Finding DNU`, and two of the three live sites do not.

---

- **`patches.txt` is the single worst merge-order hazard on this repo right now** — 12 open PRs append to the same tail (#660, #1103, #1115, #1201, #1204, #1233, #1240, #1244, #1260, #1272, #1275, #1276). All but the first to merge will conflict, and patch **order** is load-bearing here. Keep `default_finding_item` immediately after `default_gemstone_item` when you resolve.
- **`create_test_data.py`** is shared with #1115, #1201, #1260, #1275 and #1276. Whoever merges second must confirm the `Finding DNU` Item Group block and the `default_finding_item` setting value both survived the merge, or CI starts failing for reasons unrelated to their change.
- **#1260** is a 114-commit `kggk_uat → shubham_v16_aerele` branch sync whose head equals `upstream/kggk_uat`; its appearance in both rows above is an artifact of the sync, not independent work.
- In the overall suggested merge order this sits in the second group with #1251, #1204 and #1239 — small and low-collision. It is a good early merge **once the blocker is fixed**.

```mermaid
flowchart LR
  P["PR 1103 appends default_finding_item after default_gemstone_item"] --> TAIL["patches.txt tail"]
  O["11 other open PRs append to the same tail"] --> TAIL
  TAIL --> C["textual conflict for whoever merges second, and patch order is load-bearing"]
  CTD["create_test_data.py shared with 1115, 1201, 1260, 1275, 1276"] --> C2["second merger must re-check the Finding DNU group and the setting value"]
```

---

## Suggested order of work

1. **Guard the Item Group in the patch** (issue 1). This is the one that must not ship as-is — `bench migrate` currently aborts on `alfarsi` and `kg-gk`.
2. **Backfill `default_finding_item` onto existing Manufacturing Setting rows in the same patch** (issue 2), so migrate leaves every site able to submit a PMO. Do `default_gemstone_item` at the same time.
3. **Fix the pure-qty gate** (issue 3) — either resolve `variant_of` through `custom_alternative_item`, or give the dummy a real `variant_of` of `F`.
4. **Add the finding test class** to the already-registered `test_material_request_customizations.py`, plus the PMO throw test and a `custom_pure_qty > 0` assertion (issue 4).
5. Optional, but cheap while you are in the file: collapse the two validators into one helper and close the fail-open (issues 5 and 6).

Happy to go deeper on any of these — particularly the per-site Item Group and Manufacturing Setting query results behind issues 1 and 2, if it helps to see them directly.

PR #1115, opened 2026-08-20
feat: Implement Customer Gold Receipt functionality

DHINESH00:
- Added validation rules for Customer Gold receipts in the `customer_gold_receipt.py` file.
- Introduced a new doctype `Customer Gold Company Account` with necessary fields and validations.
- Enhanced `Subcontracting Settings` to include configurations for Customer Gold flow, including item and account validations.
- Updated test cases for Customer Gold receipt eligibility and validation to ensure proper functionality.
- Integrated Customer Gold validation into stock entry events to enforce rules during receipt processing.


DHINESH00:
## Review — #1115 Customer Gold Receipt functionality

&nbsp;·&nbsp; 0 blocker · 1 high · 2 medium · 4 low · 3 nit
&nbsp;·&nbsp; base `v16_develop_aerele` · head `2ea9cbc3` · 20 files, +2724/-21 (live three-dot diff: 14 files / +1673)

The headline is not a code defect: this branch is an earlier ancestor of the #1276 / #1275 Customer Gold programme, it is `mergeable=false` (dirty), and the one HIGH below is a defect the successor branch has already diagnosed and fixed in its own words. The recommendation is to close this in favour of #1276 / #1275 and carry forward only the piece they do not have. The review is still worth reading in full, because several of the findings apply to the successors too.

---

## What this PR does

Part of the API's diffstat has already landed on `upstream/v16_develop_aerele` — the `customer_gold_receipt.py` validator skeleton, `customer_gold_company_account`, and the Customer Gold block in `subcontracting_settings.py`. What is genuinely new here is four things: (1) a read-only rate service `customer_subcontracting/customer_gold_rate.py` that resolves a `Gold Rates` per-gram rate for an exact posting date; (2) a rate snapshot plus a nominal valuation inside `validate_customer_gold_receipt`, stamping `set_basic_rate_manually` / `basic_rate` / `basic_amount` / `expense_account` so the receipt posts `Dr Stock / Cr Customer Gold Liability` through ERPNext's own engine; (3) a new `Customer Gold Ledger` DocType with `on_submit` / `on_cancel` writers; (4) two `create_custom_fields` patches on Stock Entry, plus 584 lines of new tests.

A lot of this is done well, and it is worth saying which parts:

- **The GL routing is correct and was clearly reasoned about.** Stock Entry Detail has no `target_warehouse`, so `erpnext/controllers/stock_controller.py` takes the `else` branch and picks up `item_row.expense_account`; and `set_basic_rate_manually` makes `set_basic_rate` skip the `allow_zero_valuation_rate` wipe while still recomputing `basic_amount`. Debit and credit derive from the same `stock_value_difference`, with no Journal Entry anywhere. That is the right way to do this.
- **No injection surface and no new public API.** `_get_source_row` fetches `fields=["*"]` specifically to avoid interpolating the `9_am` / `3_pm` / `11_pm` column names, `rate_field` is allowlisted against `GOLD_RATE_FIELDS`, and the diff adds no `@frappe.whitelist()` at all.
- **The rate lookup is keyed on the caller's `posting_date`, never on `today()`** — the thing that quietly breaks every backdated receipt in implementations that get this wrong.
- **The liability account is re-checked for company ownership per document**, and everything sits behind `is_customer_gold_enabled()`, which fails closed.
- **Both new test modules are registered in `.github/workflows/inventory-tests.yml:67-68` in this same PR.** That is not the default behaviour on this repo — sibling PR #1240 ships 508 lines of tests with no workflow edit at all — so it is worth calling out.
- **Both patches are in `patches.txt` in the correct dependency order** (snapshot first, because `add_customer_gold_nominal_value_field` anchors `insert_after: custom_gold_rate_per_gram` on it) **and mirrored into `create_test_data.create_test_data()` in the same order**, which is the correct two-place provisioning convention on this base.

```mermaid
flowchart TD
  A["Stock Entry save, stock_entry_type = configured receipt type"] --> B["before_validate: validate_customer_gold_receipt"]
  B --> C{"is_customer_gold_enabled and type matches?"}
  C -- "no" --> Z["untouched, ordinary Stock Entry"]
  C -- "yes" --> D["customer_gold_rate.py resolves the Gold Rates row for posting_date"]
  D --> E["stamp the custom_gold_rate_* snapshot fields on the header"]
  F["set_customer_gold_nominal_valuation stamps each row:\nset_basic_rate_manually, basic_rate, basic_amount, expense_account"]
  E --> F
  F --> G["ERPNext validate: set_transfer_qty then calculate_rate_and_amount"]
  G --> H["on_submit: SLE plus GL, Dr Stock / Cr Customer Gold Liability"]
  H --> I["create_customer_gold_ledger writes one Customer Gold Ledger row per item row"]
  I --> J["on_cancel: cancel_customer_gold_ledger flips is_cancelled to 1"]
```

## Before → after

```mermaid
flowchart TD
  subgraph BEFORE["Base: upstream/v16_develop_aerele"]
    B1["Customer Gold receipt saved"] --> B2["validate_customer_gold_receipt runs the rules only"]
    B2 --> B3["no rate snapshot, no basic_rate stamped"]
    B3 --> B4["ERPNext's own valuation decides the posting value"]
    B4 --> B5["no monetary ledger row written"]
  end
  subgraph AFTER["This PR"]
    A1["Customer Gold receipt saved"] --> A2["rules, plus a Gold Rates snapshot frozen for posting_date"]
    A2 --> A3["every row stamped: set_basic_rate_manually = 1, basic_rate = per_gram,\nexpense_account = Customer Gold Liability"]
    A3 --> A4["GL posts Dr Stock / Cr Customer Gold Liability through stock_controller"]
    A4 --> A5["one Customer Gold Ledger row per item row, flipped on cancel"]
  end
```

---

## Issues at a glance

| 1 | **HIGH** | Cancel path gates on live settings, so a config change strands ledger rows at `is_cancelled = 0` | `customer_gold_ledger.py:113` | Flag switched off or SE type repointed between submit and cancel |
| 2 | **MEDIUM** | Create path is asymmetric the other way: draft-stamped valuation posts the GL with no ledger row | `customer_gold_ledger.py:61` | Flag switched off between draft save and submit |
| 3 | **MEDIUM** | The only code that writes persistent rows — the ledger writers — has no test at all | `customer_gold_ledger.py:99` | Every CI run; a green check proves nothing about the ledger |
| 4 | **LOW** | Nominal total sums stale `transfer_qty`, manufacturing the divergence the field exists to detect | `customer_gold_receipt.py:352` | Any save that changes `qty` without restating `transfer_qty` |
| 5 | **LOW** | Valuation tests assert against stubbed fakes, so a Custom Field typo still passes green | `test_customer_gold_receipt.py:96` | Any drift between the patch fieldnames and the code fieldnames |
| 6 | **LOW** | New ledger DocType has no `search_index` on any of the columns it is queried by | `customer_gold_ledger.json:88` | Progressively, as the ledger accumulates rows |
| 7 | **LOW** | Dead fallback read on every submit, plus a per-row UPDATE loop on cancel | `customer_gold_ledger.py:76` | Every Customer Gold submit and cancel |
| 8 | **NIT** | Comment justifying the explicit `basic_amount` write cites the wrong line and inverts ERPNext's behaviour | `customer_gold_receipt.py:358` | On every read by the next maintainer |
| 9 | **NIT** | Append-only audit ledger grants `delete` to System Manager | `customer_gold_ledger.json:203` | List-view bulk delete, at any time |
| 10 | **NIT** | Ships schema the successor branches replaced, with no drop patch on either side | `patches.txt:78` | If both programmes converge on one site |

---

### 1. Cancel gates on live settings, so a config change strands ledger rows as un-cancelled · **HIGH**

**What** — `cancel_customer_gold_ledger` returns early when `_receipt_settings(doc)` is falsy, and `_receipt_settings` reads TODAY's `enable_customer_gold_flow` and TODAY's `customer_goods_stock_entry_type`. Cancelling a receipt that was submitted under a since-changed configuration reverses the SLE and the GL but leaves every `Customer Gold Ledger` row at `is_cancelled = 0`.

**Why** — the two sides use asymmetric guards. The write path keys its idempotency off the rows themselves (`filters={"reference_doctype": ..., "reference_docname": ...}` at `:65-69`); the cancel path keys off live configuration at `:113`. Nothing about the already-written rows is consulted. The DocType's own `is_cancelled` description states the contract that is then broken: *"Set when the source receipt is cancelled. Rows are never deleted -- the audit trail is the point. Balance calculations must exclude rows where this is set."*

**How** — submit a receipt with the flag on: N rows land at `is_cancelled = 0` with a `nominal_value` tying to the GL credit. An admin unticks `enable_customer_gold_flow` (it ships `default: 0` and is a pilot switch) or repoints `customer_goods_stock_entry_type`. Cancel the receipt: ERPNext reverses the stock and the GL, `cancel_customer_gold_ledger` returns at `:114` without touching a row. Re-enabling the flag does not repair it — nothing re-walks cancelled documents.

**When** — any cancel that happens after the master switch is turned off or after the Stock Entry Type is edited. Both are ordinary configuration operations on a Single that any System Manager can edit.

**Where** — `jewellery_erpnext/customer_subcontracting/doctype/customer_gold_ledger/customer_gold_ledger.py:113`

One honest narrowing on blast radius: at this head nothing reads `Customer Gold Ledger` yet — `git grep -ln 'Customer Gold Ledger' pr/1115` returns only the DocType JSON, the module and `hooks.py`. So today the consequence is an audit table that permanently no longer matches the GL, rather than a wrong number on a shipped report. It is graded HIGH because the rows are wrong at write time and there is no repair path once the balance reader the module docstring promises lands.

```python
102:def cancel_customer_gold_ledger(doc, method=None):
...
109:	from jewellery_erpnext.customer_subcontracting.customer_gold_receipt import (
110:		_receipt_settings,
111:	)
112:
113:	if not _receipt_settings(doc):
114:		return
115:
116:	for name in frappe.get_all(
117:		"Customer Gold Ledger",
118:		filters={"reference_doctype": doc.doctype, "reference_docname": doc.name},
119:		pluck="name",
120:	):
121:		frappe.db.set_value("Customer Gold Ledger", name, "is_cancelled", 1)
```

```python
66:def _receipt_settings(doc):
...
71:	if doc.doctype != "Stock Entry":
72:		return None
73:	if not is_customer_gold_enabled():
74:		return None
75:
76:	settings = get_customer_gold_settings()
77:	configured_type = settings.get("customer_goods_stock_entry_type")
78:	if not configured_type or doc.get("stock_entry_type") != configured_type:
79:		return None
```

```mermaid
flowchart TD
  T1["t0: flag ON, type = the configured receipt type"] --> T2["t1: receipt submitted.\nledger rows inserted at is_cancelled = 0, GL credit posted to the liability account"]
  T2 --> T3["t2: admin unticks enable_customer_gold_flow\nor repoints customer_goods_stock_entry_type"]
  T3 --> T4["t3: receipt cancelled"]
  T4 --> T5["ERPNext reverses the SLE and both GL legs"]
  T4 --> T6["cancel_customer_gold_ledger hits the guard at :113 and returns"]
  T6 --> T7["rows stay at is_cancelled = 0. No error, no repair path,\nre-enabling the flag does not fix it"]
```

The successor branch reaches the same conclusion independently — `pr/1276 customer_gold_fulfilment.py:1550-1562`:

```python
def reverse_receipt(doc, method=None):
	"""``on_cancel`` for Stock Entry -- the mirror of :func:`record_receipt`.

	Guarded on SCHEMA, deliberately not on the feature flag ...
	No ``_receipt_settings`` check either. That reads TODAY's configured stock entry type, and a
	site that has since reconfigured it would silently fail to reverse an older receipt. The
	events themselves are the authority ...
```

**Fix** — key the reverse path off the rows, which are the only record of what the document actually did. This is also the single-UPDATE form that removes the loop in issue 7:

```python
def cancel_customer_gold_ledger(doc, method=None):
	if doc.doctype != "Stock Entry":
		return
	names = frappe.get_all(
		"Customer Gold Ledger",
		filters={
			"reference_doctype": doc.doctype,
			"reference_docname": doc.name,
			"is_cancelled": 0,
		},
		pluck="name",
	)
	if not names:
		return
	frappe.db.set_value(
		"Customer Gold Ledger",
		{"name": ("in", names)},
		"is_cancelled",
		1,
		update_modified=True,
	)
```

Keep the settings gate on `create_customer_gold_ledger` (but see issue 2), and drop the now-unused `_receipt_settings` import from this function.

---

### 2. Is the create side asymmetric the other way as well? · **MEDIUM**

**What** — `create_customer_gold_ledger` gates on `_receipt_settings(doc)` at `:61-63`, i.e. on live configuration read at **submit** time. But the monetary effect of the receipt was already persisted on the **draft**: `set_customer_gold_nominal_valuation` stamps `set_basic_rate_manually = 1`, `basic_rate` and `expense_account = <liability account>` onto every row during `before_validate` of the draft save, and those are ordinary stored Stock Entry Detail columns.

**Why** — can you confirm that a flag flip between draft save and submit does not post a liability credit with no ledger row behind it? The reading that concerns me: `validate_customer_gold_receipt` returns early at `customer_gold_receipt.py:86-88` and re-stamps nothing — but it also *clears* nothing, and the persisted row values still drive the posting. `set_basic_rate` at `erpnext/stock/doctype/stock_entry/stock_entry.py:1616-1619` recomputes `basic_amount = transfer_qty * basic_rate` for `set_basic_rate_manually` rows regardless of any jewellery flag, and the stored `expense_account` is the credit counter-leg for a Material Receipt. If that holds, the SE posts at full nominal value while `create_customer_gold_ledger` returns at `:63` and writes nothing — the exact mirror of issue 1, and not fixed by fixing the cancel path alone.

**How** — flag on, draft a receipt for 100 g at 7,000/g: rows persist with `set_basic_rate_manually = 1`, `basic_rate = 7000`, `expense_account = Customer Gold Liability`. Admin unticks the flag. Submit: the jewellery validator returns early, ERPNext's `validate()` still runs `set_basic_rate`, the GL posts Dr Stock 700,000 / Cr Customer Gold Liability 700,000, and the ledger that is supposed to reconcile to it has zero rows for the document.

**When** — any receipt drafted while the flag is on and submitted after it is turned off or after `customer_goods_stock_entry_type` is edited.

**Where** — `jewellery_erpnext/customer_subcontracting/doctype/customer_gold_ledger/customer_gold_ledger.py:61`

```python
			if d.s_warehouse or d.set_basic_rate_manually:
				if d.set_basic_rate_manually:
					d.basic_amount = flt(flt(d.transfer_qty) * flt(d.basic_rate), d.precision("basic_amount"))
				continue
```

```mermaid
flowchart TD
  D1["Draft saved with the flag ON"] --> D2["before_validate stamps each row:\nset_basic_rate_manually = 1, basic_rate, expense_account"]
  D2 --> D3["those values are persisted columns on Stock Entry Detail"]
  D3 --> D4["admin turns enable_customer_gold_flow OFF"]
  D4 --> D5["Submit"]
  D5 --> D6["validate_customer_gold_receipt returns early at :86-88.\nNothing re-stamped, nothing cleared"]
  D5 --> D7["ERPNext set_basic_rate still honours the stored\nset_basic_rate_manually row and recomputes basic_amount"]
  D7 --> D8["GL posts Dr Stock / Cr Customer Gold Liability"]
  D6 --> D9["create_customer_gold_ledger returns at :63.\nNo ledger row for a posted liability"]
```

**Fix** — gate both writers on what the document carries rather than on live settings: `doc.doctype == "Stock Entry"` plus a stamped Customer Gold marker (a non-zero `custom_gold_nominal_value`, or a row whose `expense_account` is the configured liability account), which is the "guarded on SCHEMA, deliberately not on the feature flag" approach the successor branch took. Alternatively, refuse the submit outright when the draft carries a Customer Gold valuation but the flag is now off, so the receipt cannot post value without a ledger row.

---

### 3. The ledger writers — the only new code that writes persistent rows — have no test · **MEDIUM**

**What** — `create_customer_gold_ledger` and `cancel_customer_gold_ledger` are referenced by no test anywhere in the tree at this head. There is no `test_customer_gold_ledger.py`, and no line in the two new test modules touches the ledger. The two new patches and the `_pure_qty_excluded_types()` behaviour change in `doc_events/stock_entry.py` are equally untested. The 584 lines of new tests all cover the pure-logic rate service and the in-memory validator.

**Why** — to be clear, this is *not* the usual "unregistered test never runs" criticism: this PR does register both of the test modules it ships. The gap is that the highest-risk surface has no test at all. It is the writer for a monetary ledger that must tie to the GL, it uses `ignore_permissions=True`, its idempotency rests on a single read-then-insert guard with no unique key behind it (the module docstring says so itself), and it is where issue 1 lives. A test that submitted, toggled the flag, and cancelled would have caught that defect directly. The repo already has the precedent for testing patches — `pr/1233` ships `tests/test_weight_backfill_patches.py`.

**How** — change `create_customer_gold_ledger` to insert two rows per item, or to skip the cancel flip entirely, and every workflow check still passes green, including `inventory-tests.yml`, the workflow this PR edits.

**When** — continuously, for every future change to the ledger writers.

**Where** — `jewellery_erpnext/customer_subcontracting/doctype/customer_gold_ledger/customer_gold_ledger.py:99`

```
git grep -ln 'customer_gold_ledger\|Customer Gold Ledger' pr/1115
pr/1115:jewellery_erpnext/customer_subcontracting/doctype/customer_gold_ledger/customer_gold_ledger.json
pr/1115:jewellery_erpnext/customer_subcontracting/doctype/customer_gold_ledger/customer_gold_ledger.py
pr/1115:jewellery_erpnext/hooks.py
```

```python
	for row in doc.get("items") or []:
		ledger = frappe.new_doc("Customer Gold Ledger")
		ledger.customer = row.get("customer") or doc.get("_customer")
...
		ledger.liability_account = row.get("expense_account") or liability_account
		ledger.insert(ignore_permissions=True)
```

**Fix** — the ask is a mock-level test in the same style as the two new modules (a `frappe._dict` Stock Entry stand-in plus patched `is_customer_gold_enabled` / `get_customer_gold_settings`), not a new integration suite. Add `jewellery_erpnext/jewellery_erpnext/tests/test_customer_gold_ledger.py` covering at minimum:

- one ledger row per item row, with `nominal_value` equal to `row.basic_amount` and `liability_account` equal to `row.expense_account`;
- the double-fire guard: calling `create_customer_gold_ledger` twice inserts one set of rows;
- a non-Customer-Gold Stock Entry type writes nothing;
- submit then cancel flips every row to `is_cancelled = 1`;
- cancel still flips the rows after `is_customer_gold_enabled()` is patched to `False` and after `customer_goods_stock_entry_type` is repointed — that is the regression test for issue 1.

Then register it beside the other two:

```diff
           bench --site test_site run-tests --module jewellery_erpnext.jewellery_erpnext.tests.test_customer_gold_rate
+          bench --site test_site run-tests --module jewellery_erpnext.jewellery_erpnext.tests.test_customer_gold_ledger
```

---

### 4. Nominal total sums stale `transfer_qty`, manufacturing the divergence the field exists to detect · **LOW**

**What** — `set_customer_gold_nominal_valuation` runs in `before_validate` and prefers `row.transfer_qty` over `row.qty` when deriving `stock_qty`. `transfer_qty` is not recomputed until `StockEntry.validate()` calls `set_transfer_qty()`, which runs after every `before_validate` hook. The header field is frozen from the stale value while ERPNext later recomputes each row's `basic_amount` from the fresh one.

**Why** — in the installed erpnext, `validate()` calls `set_transfer_qty()` at `:290` and `calculate_rate_and_amount()` at `:334`, and `set_basic_rate` at `:1616-1618` recomputes `d.basic_amount` from `d.transfer_qty` for exactly the `set_basic_rate_manually` rows this PR stamps. So the row amounts are always right, but `doc.custom_gold_nominal_value` is summed at line 364 from the pre-`set_transfer_qty` figures and never revisited. The `transfer_qty`-first preference also buys nothing: `_validate_row_uom` (invoked at line 347) already throws unless `conversion_factor == 1`, so `qty` IS the stock quantity by construction. And the whole justification for the extra field is tamper detection — `patches/add_customer_gold_nominal_value_field.py` says *"If the two ever disagree, something re-valued a customer receipt behind our back -- that divergence IS the audit signal"* — so a routine save that produces a divergence poisons the signal.

**How** — load a draft, set `doc.items[0].qty = 50` where `transfer_qty` is still 100, save. `flt(row.get("transfer_qty"))` is 100 and truthy, so the fallback is never evaluated: header `custom_gold_nominal_value = 700,000`. Then `set_transfer_qty()` makes it 50 and `set_basic_rate` recomputes `basic_amount = 350,000`. The document saves with a header total that is double its own rows, and nobody tampered with anything.

**When** — any save that changes `qty` without restating `transfer_qty` in the same pass: `frappe.get_doc(...).save()`, `bench execute`, a Server Script, a partial REST update, a bulk edit. Two narrowings worth stating: a fresh server-side insert is unaffected (`transfer_qty` is None, so the fallback product is used), and the divergence self-heals on the next save, because `before_validate` also runs on the submit transition and `transfer_qty` has been corrected by then. The bad value persists on a draft until it is next saved, or on a document whose qty changes in the same pass that submits it. Also, contrary to what you might expect, the desk is exposed too — there is no `qty` handler in the `frappe.ui.form.on("Stock Entry Detail")` block of `erpnext/stock/doctype/stock_entry/stock_entry.js`, only `conversion_factor` and `transfer_qty` handlers. Impact is confined to the audit-only header field; `basic_amount`, the SLE and the GL are all correct.

**Where** — `jewellery_erpnext/customer_subcontracting/customer_gold_receipt.py:352`

```python
	liability_account = _customer_gold_liability_account(doc)

	total_nominal_value = 0.0
	for row in doc.get("items") or []:
		_validate_row_uom(row)

		# transfer_qty is qty in the STOCK uom; the rate is per gram and the stock uom is
		# Gram at conversion factor 1, so the two are directly compatible. Fall back to the
		# product when set_transfer_qty has not run yet on this pass.
		stock_qty = flt(row.get("transfer_qty")) or flt(row.get("qty")) * flt(
			row.get("conversion_factor") or 1
		)

		row.set_basic_rate_manually = 1
		row.basic_rate = per_gram
		row.basic_amount = flt(stock_qty * per_gram, row.precision("basic_amount"))
		row.expense_account = liability_account

		total_nominal_value += flt(row.basic_amount)

	doc.set(
		NOMINAL_VALUE_FIELD,
		flt(total_nominal_value, doc.precision(NOMINAL_VALUE_FIELD)),
	)
```

```mermaid
flowchart TD
  S1["Loaded draft: qty edited from 100 to 50,\ntransfer_qty on the row is still 100"] --> S2["before_validate: flt(transfer_qty) is 100 and truthy,\nso stock_qty = 100"]
  S2 --> S3["row.basic_amount = 700000,\nheader custom_gold_nominal_value = 700000"]
  S3 --> S4["validate: set_transfer_qty makes transfer_qty 50"]
  S4 --> S5["calculate_rate_and_amount: set_basic_rate\nrecomputes basic_amount = 350000"]
  S5 --> S6["saved: header 700000 against rows totalling 350000.\nA false re-valued-behind-our-back signal"]
```

**Fix** — stop preferring the stale field; `_validate_row_uom` already guarantees `conversion_factor == 1`:

```diff
-		stock_qty = flt(row.get("transfer_qty")) or flt(row.get("qty")) * flt(
-			row.get("conversion_factor") or 1
-		)
+		# conversion_factor is forced to 1 by _validate_row_uom above, so qty IS the stock
+		# quantity. transfer_qty must NOT be used here: set_transfer_qty() has not run yet on
+		# this pass and the value on the row may still be the previously saved one.
+		stock_qty = flt(row.get("qty"))
```

If you would rather the header tie to whatever ERPNext finally books, move only the `doc.set(NOMINAL_VALUE_FIELD, ...)` line into a Stock Entry `validate` hook (which runs after the controller's own `calculate_rate_and_amount()`) and sum `row.basic_amount` there. Either way, a regression test that changes `qty` on a loaded draft and asserts `custom_gold_nominal_value == sum(row.basic_amount)` after save.

---

### 5. Valuation tests assert only against stubbed fakes, so a Custom Field typo still passes green · **LOW**

**What** — every assertion added to `test_customer_gold_receipt.py` runs against a `frappe._dict` with hand-written stand-ins for the two `Document` methods the production code calls. No test touches a real `Stock Entry`, and the module docstring concedes it: *"Stock Entries are `frappe._dict` fakes and every DB read is patched. No document is created."*

**Why** — the stubbed `doc.set` always stores the key; a real `Document` does not. `frappe/model/base_document.py:510-519` builds the row from `self.meta.get_valid_fields()`, so an attribute whose Custom Field was never provisioned is silently dropped at write time — no exception, no warning. And the eight fields this feature depends on (`custom_gold_nominal_value` plus the seven `custom_gold_rate_*`) exist only because the two patches ran; `hooks.py:12` has `after_migrate` commented out, so `custom_fields/*.json` is inert. The tests would pass identically if a fieldname in `RATE_SNAPSHOT_FIELDS` were misspelled in one place and not the other.

In fairness: `create_test_data.py` gains 18 lines calling both patches' `execute()` in order, so CI's `test_site` and fresh sites do get the fields provisioned. The surviving point is narrower than "the patch might not have run" — it is fieldname-drift blindness.

**How** — typo `custom_gold_rate_per_gram` or `custom_gold_nominal_value` in either the patch or `customer_gold_receipt.py` but not the other, and all four workflow jobs stay green. In production the value is discarded at write time, the field stores 0, and the audit-divergence signal the patch docstring is built around is destroyed rather than raised.

**When** — any drift between patch fieldnames and code fieldnames.

**Where** — `jewellery_erpnext/jewellery_erpnext/tests/test_customer_gold_receipt.py:96`

```python
77:	row.precision = lambda fieldname: 2
...
94:	doc.precision = lambda fieldname: 2
95:	# frappe._dict.set() exists, but the real Document.set is what the code calls.
96:	doc.set = lambda fieldname, value: doc.update({fieldname: value})
```

**Fix** — keep the fast mock classes and add one cheap guard that costs nothing and fails loudly on drift:

```python
def test_snapshot_fields_exist_on_stock_entry(self):
	meta = frappe.get_meta("Stock Entry")
	for fieldname in (*RATE_SNAPSHOT_FIELDS, NOMINAL_VALUE_FIELD):
		self.assertTrue(meta.has_field(fieldname), fieldname)
```

If you want one end-to-end case as well, a single `IntegrationTestCase` class that submits a real receipt on `test_site` and asserts `frappe.db.get_value("Stock Entry", name, "custom_gold_nominal_value")` and the resulting SLE `stock_value_difference` would close the gap — that class must call `super().setUpClass()`, since a bare `pass` override disables the rollback cleanup.

---

### 6. New ledger DocType carries no index on the columns it is queried by · **LOW**

**What** — `customer_gold_ledger.json` declares 14 fields and not one sets `search_index`. Frappe emits a secondary index only for a field that sets it (`frappe/database/schema.py:105` is the only place `set_index` comes from, and there is no auto-indexing of Link or Dynamic Link fields), so `tabCustomer Gold Ledger` ships with no index on `reference_docname`, `reference_doctype`, `customer`, `company` or `is_cancelled`.

**Why** — those are exactly the columns the module's two access paths filter on, and the columns the documented future readers will use (*"Day 6's balance calculation reads this; so, later, do allocation and delivery settlement"*). This is append-only by design and never pruned, so it only grows.

Two things that keep this at LOW rather than higher: the master switch ships `default: 0`, so the table is empty on every site today; and the incumbent ledger this module compares itself to, `Subcontracting Log`, likewise declares zero `search_index` fields. So this is house scaffold, not a regression — it is a cheap schema-hygiene fix to make before the ledger accumulates, rather than a cost anyone is paying now.

**When** — progressively, once the feature is piloted.

**Where** — `jewellery_erpnext/customer_subcontracting/doctype/customer_gold_ledger/customer_gold_ledger.json:88`

```json
  {
   "fieldname": "reference_docname",
   "fieldtype": "Dynamic Link",
   "options": "reference_doctype",
   "label": "Reference Document",
   "reqd": 1,
   "read_only": 1,
```

**Fix** — add `"search_index": 1` to `reference_docname`, `customer` and `company`. Separately, the module docstring notes *"there is no unique key on `(reference_docname, batch)`"*, which leaves the `:65-74` read-then-insert as the only thing between a double-fire and duplicated financial rows; if you want that to be a real constraint rather than a best-effort check, add a composite unique index in the provisioning patch over `(reference_doctype, reference_docname, item, batch)`.

---

### 7. Dead fallback read on every submit, and a per-row UPDATE loop on cancel · **LOW**

**What** — two small query-shape issues. (1) `create_customer_gold_ledger` runs a `frappe.db.get_value("Stock Entry Detail", ...)` unconditionally at `:76-78`, purely as a per-row fallback that is unreachable in practice. (2) `cancel_customer_gold_ledger` issues one `frappe.db.set_value` per row at `:116-121` where a single filtered UPDATE would do.

**Why** — `set_customer_gold_nominal_valuation` stamps `row.expense_account = liability_account` on every row, and `_customer_gold_liability_account` throws rather than returning empty, so `row.get("expense_account")` at `:98` is always truthy and the query's result is always discarded. Nothing in ERPNext clears it afterwards either — `stock_entry.py:843-854` only fills `expense_account` when falsy, and the `:863-864` override applies to `Subcontracting Delivery`, not a Material Receipt. If the fallback ever did fire, a dict-filter `get_value` defaults to ordering by `creation`, so it would pin the liability account of the **earliest-created** item row, which is still the wrong row to pin it to.

**When** — every Customer Gold submit and cancel once the flag is on. Nothing is corrupted today; the dead fallback is a latent hazard if the valuation step is ever made conditional, as the successors did with their `valuation_policy` gate.

**Where** — `jewellery_erpnext/customer_subcontracting/doctype/customer_gold_ledger/customer_gold_ledger.py:76`

```python
76:	liability_account = frappe.db.get_value(
77:		"Stock Entry Detail", {"parent": doc.name}, "expense_account"
78:	)
...
98:		ledger.liability_account = row.get("expense_account") or liability_account
```

**Fix** — delete the fallback read, since `row.expense_account` is guaranteed:

```diff
-	liability_account = frappe.db.get_value(
-		"Stock Entry Detail", {"parent": doc.name}, "expense_account"
-	)
...
-		ledger.liability_account = row.get("expense_account") or liability_account
+		ledger.liability_account = row.get("expense_account")
```

The cancel loop collapses into the single statement already shown in issue 1's fix.

---

### 8. Does the comment at `:358-360` describe ERPNext correctly? · **NIT**

**What** — the comment says the explicit `row.basic_amount` assignment *"MUST be set explicitly"* because *"the same `continue` that protects basic_rate also skips ERPNext's own basic_amount computation (`stock_entry.py:1485`)"*. Can you confirm that against the erpnext this bench runs (16.34.1)? Reading `stock_entry.py:1616-1619` there, `basic_amount` is computed for exactly these rows *before* the `continue`, not skipped by it — and `:1483-1492` is a fragment of the "Cannot transfer ... Maximum transferable quantity" throw, unrelated to rate calculation. If that reading is right, the assignment at `:361` is a same-request convenience rather than load-bearing, and it is the redundant half that reads the stale `transfer_qty` behind issue 4.

**Where** — `jewellery_erpnext/customer_subcontracting/customer_gold_receipt.py:358`

**Fix** — either drop the redundant assignment (which also resolves issue 4), or reword the comment to say that `set_basic_rate` recomputes the same figure at validate. Either way, pin the erpnext version the citation refers to — the bench is on 16.34.1 while the repo docs still say 16.25.0, so a bare line number ages badly.

---

### 9. Append-only ledger grants `delete` to System Manager · **NIT**

**What** — the new DocType grants `"delete": 1` to System Manager while its own `is_cancelled` description and `customer_gold_ledger.py:103` both state that rows are never deleted because the audit trail is the point. Nothing in code deletes rows; the permission is the path.

**Why** — a deleted row leaves no `is_cancelled` marker and no reversing document, so the tie to the GL credit becomes unverifiable. Worth being precise about what declining `delete` buys, though: a System Manager holds Role Permission Manager and can re-grant it in two clicks, and server scripts are enabled bench-wide, so `frappe.delete_doc(..., ignore_permissions=True)` bypasses the DocPerm entirely. It is a guard against an accidental list-view bulk delete, not an enforceable append-only control. And the incumbent `Subcontracting Log` ships the byte-identical permission block, so this is the repo's scaffold rather than something this PR introduced — take it as an improvement for the new ledger.

**Where** — `jewellery_erpnext/customer_subcontracting/doctype/customer_gold_ledger/customer_gold_ledger.json:203`

```json
 "permissions": [
  {
   "create": 1,
   "delete": 1,
   "email": 1,
   "export": 1,
   "print": 1,
   "read": 1,
   "report": 1,
   "role": "System Manager",
   "share": 1,
   "write": 1
  },
```

**Fix** — drop `delete` (and `write`, since every field is `read_only: 1` and the rows are only ever written by `insert()` / `set_value`). If an escape hatch is wanted for a botched pilot, expose it as an explicit maintenance patch rather than a standing role permission.

---

### 10. Which ledger design wins? · **NIT**

**What** — this PR creates a `Customer Gold Ledger` DocType and a `Stock Entry.custom_gold_nominal_value` Custom Field via `patches.txt:77-78`. Neither exists in the successor Customer Gold branches: #1276 and #1275 ship `Customer Gold Ledger Entry` and `Customer Gold Company Account` instead, their `patches.txt` registers only `add_customer_gold_rate_snapshot_fields`, and `git grep -n 'custom_gold_nominal_value' pr/1276` returns nothing at all. They also gate nominal valuation behind a `valuation_policy` switch (`VALUATION_NOMINAL` at `pr/1276 customer_gold_receipt.py:37,266`) that does not exist on this base — this PR values unconditionally.

This is merge coordination, not a defect in this diff — there is nothing #1115 could ship to fix it on its own, since a patch dropping the DocType it just created would be incoherent. And the three PRs target three diverged bases (#1115 on `v16_develop_aerele`, #1276 on `kggk_uat`, #1275 on `kggk_prod`), so the two-ledger collision needs those branches to converge on one site rather than following automatically. But if they do converge, migrated sites end up carrying a dead `tabCustomer Gold Ledger` alongside the live `Customer Gold Ledger Entry`, plus a `custom_gold_nominal_value` that stops being written and freezes at a stale total — inverting the audit signal the patch docstring exists to provide.

**Where** — `jewellery_erpnext/patches.txt:78`

```
77:jewellery_erpnext.patches.add_customer_gold_rate_snapshot_fields
78:jewellery_erpnext.patches.add_customer_gold_nominal_value_field
```

**Fix** — decide which ledger model wins before this propagates. If it is #1276 / #1275, close this PR and carry across only what they lack. If any part of #1115 must land, add a drop patch on the successor side that removes the orphaned DocType and Custom Field after migrating any rows.

---

## Tests & CI

- **Both new test modules are registered in this PR.** `.github/workflows/inventory-tests.yml:67-68` carries `test_customer_gold_receipt` and `test_customer_gold_rate`. Given that CI is a curated allowlist here, this is the right behaviour and it is not the norm on this batch of PRs.
- **`test_customer_gold_rate.py` is good negative coverage** — thirteen distinct blocking cases (missing date, missing document, missing and duplicate source row, null / zero / negative rate, blank and arbitrary settings, missing DocType).
- **The `setUpClass: pass` overrides are safe here.** Every class in both modules is mock-only — `frappe._dict` stand-ins, every DB read patched, no document created — so disabling the class rollback leaks nothing. Neither module calls `frappe.db.commit()`.
- **Nothing tests anything that writes a row.** No `test_customer_gold_ledger.py` exists (issue 3); neither patch has a test, despite `pr/1233` establishing the pattern with `test_weight_backfill_patches.py`; and the `_pure_qty_excluded_types()` change in `doc_events/stock_entry.py`, which alters `custom_pure_qty` behaviour for the configured receipt type on every Stock Entry save, is untested.
- **The tests that do exist cannot fail on a Custom Field problem** (issue 5), because the stubbed `doc.set` accepts any key while a real `Document` silently drops unknown fields.

- **#1276 (`kggk_uat`) and #1275 (`kggk_prod`)** — this PR is their ancestor. It shares `customer_gold_receipt.py`, `customer_gold_rate.py`, `patches/add_customer_gold_rate_snapshot_fields.py`, `test_customer_gold_rate.py` and `test_customer_gold_receipt.py` with both. They rewrite `customer_gold_receipt.py` (+259/-170 against this head), extend `customer_gold_rate.py`, drop `add_customer_gold_nominal_value_field.py` entirely, and replace `Customer Gold Ledger` with `Customer Gold Ledger Entry` written by `customer_gold_fulfilment.record_receipt` / `reverse_receipt`. They also fix issue 1 explicitly, in a docstring that describes exactly the defect reported above.
- **`patches.txt` is contended by 12 open PRs** (#660 #1103 #1115 #1201 #1204 #1233 #1240 #1244 #1260 #1272 #1275 #1276), all appending to the same tail. Patch order is load-bearing here: `add_customer_gold_nominal_value_field` must stay after `add_customer_gold_rate_snapshot_fields`, because it anchors `insert_after: custom_gold_rate_per_gram`. Re-check that ordering after every rebase.
- Other shared files worth knowing about: `hooks.py` (10 PRs), `doc_events/stock_entry.py` (7), `create_test_data.py` (6), `.github/workflows/inventory-tests.yml` (6).

```mermaid
flowchart LR
  P1115["PR 1115 on v16_develop_aerele:\nCustomer Gold Ledger + custom_gold_nominal_value"]
  SHARED["shared: customer_gold_receipt.py, customer_gold_rate.py,\nadd_customer_gold_rate_snapshot_fields.py, both test modules"]
  P1276["PR 1276 on kggk_uat:\nCustomer Gold Ledger Entry + valuation_policy"]
  P1275["PR 1275 on kggk_prod:\nCustomer Gold Ledger Entry + valuation_policy"]
  PT["patches.txt tail: 12 open PRs append here,\norder is load-bearing"]
  P1115 --- SHARED
  SHARED --- P1276
  SHARED --- P1275
  P1115 --- PT
  P1276 --- PT
  P1275 --- PT
```

## Suggested order of work

1. **Decide whether this branch lives.** It is dirty against `v16_develop_aerele`, part of it has already landed there, and #1276 / #1275 carry the same programme further. The cheapest path is to close this in favour of them and port across only the one thing they do not have — `custom_gold_nominal_value` as a tamper-detection field, if it is still wanted (and if so, fix issue 4 first, or the signal is noisy from day one).
2. **If it lives:** fix issue 1 — the cancel path must key off the rows, not live settings. Port the same fix into #1276 / #1275 regardless of what happens here; they inherit the question in issue 2 even though they solved issue 1.
3. Answer issue 2, and make both writers gate on document state rather than live configuration.
4. Add `test_customer_gold_ledger.py` with the config-drift cancel case, register it in `inventory-tests.yml`, and add the `meta.has_field` guard from issue 5.
5. Fix issue 4 (`stock_qty` from `qty`), then the cleanups: issues 6, 7, 8, 9.

Happy to go deeper on any of these — particularly issue 2, where I would rather be shown I have misread the draft-to-submit window than have it confirmed.

